PR #9, opened 2015-09-22
Use absolute paths in #line directives generated

matthijskooijman:
Multiple .ino files in a sketch are concatenated together, adding #line
directives so error messages refer to the original filenames. However,
these directives used plain filenames, without a path. Since these
filenames end up in the debug info as-is, this complicates using a
debugger on the resulting .elf file. Using full pathnames fixes this.

This fixes arduino/Arduino#3746.

### DIFF
--- a/src/arduino.cc/builder/sketch_source_merger.go
+++ b/src/arduino.cc/builder/sketch_source_merger.go
@@ -32,7 +32,6 @@ package builder
 import (
 	"arduino.cc/builder/constants"
 	"arduino.cc/builder/types"
-	"path/filepath"
 )
 
 type SketchSourceMerger struct{}
@@ -52,7 +51,7 @@ func (s *SketchSourceMerger) Run(context map[string]interface{}) error {
 }
 
 func addPreprocLine(sketch *types.SketchFile) string {
-	source := "#line 1 \"" + filepath.Base(sketch.Name) + "\"\n"
+	source := "#line 1 \"" + sketch.Name + "\"\n"
 	source += sketch.Source
 	source += "\n"
 
--- a/src/arduino.cc/builder/test/coan_runner_test.go
+++ b/src/arduino.cc/builder/test/coan_runner_test.go
@@ -34,7 +34,6 @@ import (
 	"arduino.cc/builder/constants"
 	"arduino.cc/builder/types"
 	"github.com/stretchr/testify/require"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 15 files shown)


ffissore:
TestMergeSketch (file sketch_source_merger_test.go) now fails. Since it has to pass on different PCs and OSs, you should add a placeholder to merged_sketch.txt and replace it at test time using the actual absolute path sketch1.ino

matthijskooijman:
@ffissore, is there any test already with similar placeholders to copy from / keep consistency with?

ffissore:
None. However there's an Abs helper func that you can use
Il 22/set/2015 17:40, "Matthijs Kooijman" <notifications@github.com> ha
scritto:

matthijskooijman:
I pushed one more commit to fix the testcase. I ended up using the text/template package to allow interpolating anything from the context, which is (more than) sufficient for this usecase (and easier than implementing some string replace manually). If you think it's ok, I'll squash the commits together to prevent breaking the testcases in the first commit.

matthijskooijman:
Oh, seems there's more testcases breaking, hold on.

matthijskooijman:
Ok, pushed a few more fixes, all tests succeed again now.

ffissore:
Uhm are you sure? All tests from prototypes_adder_test.go are failing now

matthijskooijman:
Seems I did just `git commit --amend` and forgot the `-a` :-)

Fixed now.

ffissore:
:+1:

ffissore:
Sorry @matthijskooijman, after I merged it I noticed your second commit was not signed off. I can't reopen this PR. Please make another one and ensure each of your commits is signed off. Might be worth installing this local git hook https://github.com/arduino/arduino-builder/wiki/Be-sure-your-commit-is-Signed-off-by

matthijskooijman:
W00ps! I'll submit a new PR with signoff in a minute.

Perhaps it's also an idea to have github automatically check PRs for signoff lines? Seems github itself cannot do this, but it should be easy to use a webhook for this ([see this example](https://github.com/arduino/Arduino/issues/3745#issuecomment-142365575)), or perhaps this can be added to ArduinoBot?

ffissore:
Yes, that's part of an infrastructure work I plan to do next week

matthijskooijman:
Btw, I just realized I didn't sign off the commit because I wanted to squash it with the first commit (I left it as separate commits for easier review first). Also, this commit does have some side effects (like breaking errror highlighting and making the error display less clear, see https://github.com/arduino/Arduino/issues/3745#issuecomment-142365575), so some additional changes are needed in the IDE too. Let's track those in that issue, I'll  re-open it.

I've also modified to hook in the wiki page to only give a warning about improper messages, without actually breaking the commit. This still allows committing things for testing without having to sign it off, or allows locally commiting other people's code that you can't sign off. Requiring signoff for all commits will actually reduce its value, since people will then just signoff everything, even if not appropriate.

PR #87, opened 2015-12-29
Modularization of ctags and some refactorings

cmaglie:
- "ctags" is now a module and allows an easier exchange of "parsing engines"
- added an explicit PropertiesMap type that simplify the use of `map[string]string` in most cases and
remove redundant utility functions
- other minor cleanups

/cc @matthijskooijman @facchinm


matthijskooijman:
Looks good to me!

PR #92, opened 2016-01-08
Rename ide-version to core-api-version

matteosuppo:


cmaglie:
I've merged your commit with the renamed parameter, but re-added and deprecated the old one to keep compatibily, we'll remove it in the next major release.

PR #93, opened 2016-01-14
Add the warning tag to the missing build board msg

matteosuppo:
It's a message that already has the Warning Log Level.


00alis:
@mastrolinux needs to implement Drone to speed up the testing progress :)

PR #95, opened 2016-01-15
Add the warning tag to the missing build board msg

matteosuppo:
Like https://github.com/arduino/arduino-builder/pull/93

but with the signoff
Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>


matthijskooijman:
Ah, good point. For future reference: You can update an existing pullrequest by doing a force push to its branch.

PR #96, opened 2016-01-15
Added test for building all problematic sketches

cmaglie:
- added a mode that shows the preprocessing result
- fixed some wrong sketches in the test-suite (they were good to test preprocessing but failing when building)
- added the TryBuildXXX tests

/cc @matthijskooijman


matthijskooijman:
AFAIU, this just builds all the existing sketches and checks that they compile succesfully (in addition to the more specific tests that were already happening with them)? And this does not add any sketches (yet) that are known to fail with the current code?

I've looked over the commits, they look good to me (though I haven't looked very thoroughly).

cmaglie:
> AFAIU, this just builds all the existing sketches and checks that they compile succesfully (in addition to the more specific tests that were already happening with them)? And this does not add any sketches (yet) that are known to fail with the current code?

exactly, will add more sketch after merging.

PR #104, opened 2016-01-22
Use godep to manage dependencies - fix #101

### LINKED ISSUE #101: Use godep to manage dependencies
Would be good to use godep to manage the dependencies of the project :)

matjack1:
I've been struggling to use godep in this project because godeps expects the project to comply to the [standard workspace organisation](https://golang.org/doc/code.html#Workspaces) which is slightly different from the current one.

So in this PR I propose a reorganization of the code in the first commit and the management of godep dependencies in the second one.

Reading about godep, I've seen that in Go 1.5 there is a new support for vendoring, that might be interesting. BTW, I've used Go 1.5.3 (the current stable) to run godeps, do you prefer to use the 1.4.3?

I see that this PR is quite big, so I'm not sure if you are interested in doing that, but I couldn't find a simpler way to introduce godep. I would be happy to adapt the PR if you have any suggestion.


matjack1:
I forgot to update the README.md also to better explain how the folder structure is expected to be by godep. It will probably make clearer also the PR.

matthijskooijman:
I'm not so sure I like the godeps approach of storing all dependencies (including their source code) in this git repository. I guess that storing the source inside the arduino-builder directory on your system is ok, since it makes the directory more contained, though also a bit more complicated.

Can you expand a bit more on the advantages of this approach?

matjack1:
Thanks a lot for your feedback @matthijskooijman!

Let me try to reply to both your concerns:

1. About Godep.

  I've basically followed [their reccomendation](https://github.com/tools/godep#how-to-use-godep-with-a-new-project), which says:

  > Then commit the whole Godeps directory to version control, **including Godeps/_workspace**

  I agree with you in avoiding to store the `_workspace` folder, but I've followed the Godep guidelines, on which you can read more on [StackOverflow](http://stackoverflow.com/questions/26334220/should-i-commit-godeps-workspace-or-is-godeps-json-enough). I think we can decide. My opinion is to stick to the community guidelines, which now are still committin everything in the repo (this is what looks like to me at least).

2. About the folder structure.

  I agree with you about the redundancy of the naming, but unfortunately I think that arduino-builder needs to be part of the naming, so we can't really use simply builder, unless we change the repo name. I tried to move all the content in the root, but then we have a conflict with the main package, so again I don't know if we can really use either just "builder" or just "arduino-builder". I'm happy to try more stuff if you have suggestions.

  Regarding the commit message instead, I thought having explained the folder structure in the README file, have you seen it, what do you think? Shall I add something also in the commit message?

Thanks again for reviewing the code!

matteosuppo:
Are you familiar with the [go vendor flag](https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit)?

It will probably be a better idea for long term purposes.

Also There's no need to change the imports when using godep. You can setup your workspace, run ```godep restore``` and the workspace is populated with the packages that were saved in the _workspace directory

matjack1:
Yes @matteosuppo I've seen the vendor flag, do you think that would be better? Isn't that a problem that it's only supported in Go 1.5? If not I can try to open a new PR with that and we can compare the two.

Instead I'm not sure having understood what you mean about imports. Are you talking about the dependencies stored in the _workspace folder?

matjack1:
Are you interested in me trying to use the new vendor approach: https://golang.org/doc/go1.6#go_command ? It's enabled by default with Go 1.6 https://blog.golang.org/go1.6

matteosuppo:
Well I was going to do it myself :P but first we must make sure the builder
works well with Golang 1.6 (there shouldn't be problems)

Il giorno gio 18 feb 2016, 10:49 Matteo Giaccone <notifications@github.com>
ha scritto:

cmaglie:
We are now using the official `go.mod` for dependency tracking.

PR #113, opened 2016-02-08
Use a more standard folder architecture

matteosuppo:
We follow this structure: https://golang.org/doc/code.html#Workspaces

This could be a step toward #101


facchinm:
Rebased and superseded by #116

PR #117, opened 2016-02-12
One way to handle multi-line prototypes in arduino files

focalintent:
A somewhat heavy handed way to deal with arduino/arduino-builder#80  if we detect that the code for a tag is multiline (that is, there's no closing paren), use the Filename/Line info in the tag to attempt to read a few lines from the original code to hopefully get a better block of code to look for the prototype in.

### DIFF
--- a/src/arduino.cc/builder/ctags/ctags_parser.go
+++ b/src/arduino.cc/builder/ctags/ctags_parser.go
@@ -38,6 +38,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"bufio"
 )
 
 const KIND_PROTOTYPE = "prototype"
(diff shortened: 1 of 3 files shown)


focalintent:
While this does add a bunch of file i/o from a quick run of a few dozen apps here, it appears that most multi-line prototypes end up coming out of ino files, and the time overhead for this doesn't appear to be noticeable in my runs here.

facchinm:
Hi @focalintent , this is a great news! Would you mind adding a test under ctags_parser_test.go to demonstrate the new functionality?
Thanks a lot!

focalintent:
Sure - it'll take a little bit for me to get one put together (first time messing with go at all - and also, because this fix involves reaching back into the original file mentioned in the file field of the ctags output to open it up to do the prototype finding/matching, I can't quite use the existing ctags parser test templates - though I think i'm getting a rough idea how to put together a test for it).

focalintent:
Doing a test for this is a bit more complex - because the fix involves going back to the original input file the existing test case's plan of reading text from lines of ctags output isn't sufficient to test, and unfortunately, I'm about to take off for a couple of weeks, so I won't be able to get back to this until mid-march some time.

focalintent:
Ah, saw how to do sketch/build tests - should be able to do that later tonight.

focalintent:
@facchinm added a test case for this (sorry for the delay!  new to go and the arduino-builder source)

facchinm:
Rebased and merged as fbfaa5603aa984b88d2a295658a1a57d57415b3a

PR #118, opened 2016-02-23
Add test for templates containing shifts

facchinm:
Ctags is failing badly with this case (encountered on FastLeds, `samd` architecture)


facchinm:
This must pass to close #120 as solved

PR #29, opened 2015-10-05
Adding support for sketch-specific build properties

stevemarple:
Tests for existence of a build_props.txt file in the sketch
directory. If present reads properties similar to those found in the
hardware and board files, allowing the user to easily customise the
build for a specific sketch. Most useful when containing lines similar
to

```
compiler.c.extra_flags=-D NDEBUG
compiler.cpp.extra_flags=-D NDEBUG -D MYLIBRARY_BUFSIZE=100
```

which enables the user to have macros defined for all compilation
units, including libraries.

Signed-off-by: Steve Marple stevemarple@googlemail.com

PR #119, opened 2016-02-23
Add build.project_path to global properties

robertinant:
This patch adds the build.project_path to the global properties.
One of my tools that I run as a recipe.hooks.sketch.prebuild step needs access to the path the Sketch is located at. Would be great if you could pull this!


robertinant:
@facchinm, any chance this can get pulled in into the next version?

Thanks!

Robert

facchinm:
Hi Robert,
I believe that your patch can be merged, and I'd merge it, but I need to know what you are going to do with that extra path, because if it's a way to revamp #29 I'm quite confident that it won't be well accepted (as all the previous proposals to introduce this feature)

robertinant:
My excuses for the delay in replying to this.

I wan't aware of #29 and my request is unrelated. What I am after is that when the Sketch is unsaved (e.g. I open an example and hit verify), the unsaved Sketch is written to a tmp dir. Currently arduino builder does not provide a way to know what that path is. My preprocessing tool (ino2cpp.jar) needs access to that path so that I can run the preprocessor on the Sketch.

Thanks!

Robert

robertinant:
@facchinm any chance that this can get pulled into the next version? This will help me a lot with not having to maintain my own version of the builder.

Thanks!

Robert

PaulStoffregen:
If merged, this almost certainly will open the door to #29

robertinant:
I am not understanding what #29 tries to do. The Sketch folder, if the Sketch is saved, can be figured out without issues. The issue for me is that if the Sketch is not saved, arduino-builder saves it to a temporary folder which is not exposed and thus inaccessible. So the only thing I am asking for is that the Sketch folder is know in the unsaved case which is no different from the saved case.

Hope this helps.

robertinant:
Any chance this can still be pulled in? If not then please do close this pull request. The unfortunate result of not pulling this in will be that some of the targets currently available in Energia will not work in the Arduino IDE. The reason for this pull request was to enable the Arduino IDE to pull in those Texas Instruments targets that depend on this pull request. If not pulled in, then that's OK really, I'll just disable those targets in the json file consumed by the "Additional Boards Manager URL's".

Thanks for considering this pull request. Really appreciate it!

As an off topic note. Thank you for the awesome work on arduino-builder! You did a great job pulling the build system out of the IDE with arduino-builder!

Robert

Walther:
Hey! Sorry for bumping such an old PR, but this would still be nice to have, as commented above. (I came here from https://github.com/Microsoft/vscode-arduino/issues/355 )

PR #124, opened 2016-03-04
fix multitab error reporting

facchinm:
Fixes https://github.com/arduino/arduino-builder/issues/99

Simply add the filename to any #line directive


facchinm:
Adding test right now
@per1234 @matthijskooijman

PR #125, opened 2016-03-04
Use CPP recipe to generate PREPROC if empty

facchinm:
Since the `preproc` rule is nothing more than the `cpp` rule with some extra flags and a different target, generate it instead than manually compiling a list of "compatibility_flags" and relying on it

solves #106
@matthijskooijman


facchinm:
@ArduinoBot build this please

facchinm:
Merging this PR to let it stay in staging for some weeks, eventually reverting it if testers find any drawback

Issue #86, opened 2015-12-29
Improve handling of sketch source file handling


matthijskooijman:
Prompted by https://github.com/arduino/Arduino/issues/4335 I noticed that the way that sketch source files are handled in arduino-builder should probably be improved. There's a few sides to this:
 - Instead of accepting a single .ino file, arduino-builder should probably also accept the sketch directory (since you're compiling the entire sketch, not just he .ino file passed). In general, it's probably the most flexible to accept either the directory, or any file inside (not limited to the .ino file).
 - In addition to the main sketch, it should accept one (or perhaps any number) of "override" directories. Files in these override directories are used in favor of the original ones, but any files missing from the override directories are just used from the original sketch directory. This is useful when the IDE files have unsaved changes, which now cause the entire sketch to be written to a temporary directory and that sketch is compiled instead. This is wasteful (since also unchanged files are copied) and prevents arduino-builder from properly detecting wether a file was changed (this is currently solved by doing a contents comparison, but timestamps are probably better).
 - Instead of doing content comparison for files to decide if they changed, we should probably just check timestamps. The previous change allows this, since only changed files are written.
 - When actually compiling the sketch, arduino-builder compiles any source files present in the build directory (which are copied there, because they need to be in the same directory as the preprocessed .ino files). However, arduino-builder does not currently remove any source files no longer present in the original sketch. This means that deleting a file from the sketch causes it to still be compiled by arduino-builder. The IDE has some code to delete source files from the build directory when a file is deleted through the IDE, but this is really the wrong place to solve this. To properly solve this, either arduino-builder should delete any "extra" files when starting a build (taking care of keeping the .o files around) or it should just ignore the extra files (e.g. compile only files present in the original sketch, instead of listing files in the build directory). I tend towards the latter option, since the former will also delete any extra files generated by the commands in platform.txt, which prevents a third-party platform.txt from doing some things, like keeping a cache of sorts between compile runs). The downside of the latter approach is that it keeps extra files around in the build directory, but that should only be until the IDE is closed, so  I don't think it's a big problem.
 - When copying sketch source files into the build directory for building, arduino-builder should probably add `#line` directives, so the compiler knows the original filename (which can help with error messages and debugging symbols).

Some of these might be somewhat separate, but I mostly wanted to get these thoughts out of my head so I won't forget about them later :-)

matthijskooijman:
[...]

PR #129, opened 2016-03-22
Don't wipe everything is only sketch path changed

facchinm:
If the environment variables are untouched but the sketch is saved automatically by the Java IDE (to another folder) the infamous "Build options changed" is triggered.
This commits adds a second check after removing the sketch location from the env variables, and skips the wipeout if the partial environments are equals

Should solve https://github.com/arduino/Arduino/issues/4713
@matthijskooijman @cmaglie @sandeepmistry @mbanzi

### DIFF
--- a/src/arduino.cc/builder/wipeout_build_path_if_build_options_changed.go
+++ b/src/arduino.cc/builder/wipeout_build_path_if_build_options_changed.go
@@ -36,6 +36,7 @@ import (
 	"arduino.cc/builder/utils"
 	"os"
 	"path/filepath"
+	"regexp"
 )
 
 type WipeoutBuildPathIfBuildOptionsChanged struct{}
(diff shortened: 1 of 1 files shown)


facchinm:
@ArduinoBot build this please

matthijskooijman:
Wouldn't this cause cache to be re-used even when you load a different sketch? Or are there other variables that will change then as well?

In any case, I think the proper fix for this problem is to let arduino-builder know both the original sketch location as well as the temporary saved copy, which allows it to be smarter about other things (like debugging symbols) too. See #86 for a writeup on how I think this should look.

facchinm:
Changing the IDE active window or changing the target board triggers a wipe, so there shouldn't be much issues. Anyway, I'm merging this to give it some staging time in nightly and eventually, if there is any drawback, we'll take care of reverting

PR #131, opened 2016-04-07
Avoid using whole ctags_target_for_gcc_minus_e for prototype generation

facchinm:
Ctags is FULL of bugs on c++ parsing (and c++ syntax is a nightmare), sketches are usually way more simple (and easier to debug).
Using a simplified version of the ctags_target_for_gcc_minus_e file which only contains preprocessor directives and the sketch lines helps avoiding most ctags subtle bugs.

### DIFF
--- a/src/arduino.cc/builder/ctags_target_file_saver.go
+++ b/src/arduino.cc/builder/ctags_target_file_saver.go
@@ -79,10 +79,10 @@ func saveLinesContainingDirectivesAndSketch(src string, tofind []string) string 
 		if saveLine || startsWithHashtag(line) {
 			minimizedString += line + "\n"
 		}
-		if containsAny(line, tofind) && isPreprocessorLineMarker(line) {
+		if containsAny(line, tofind) && isLineMarker(line) {
 			saveLine = true
 		}
-		if saveLine && !containsAny(line, tofind) && isPreprocessorLineMarker(line) {
+		if saveLine && !containsAny(line, tofind) && isLineMarker(line) {
 			saveLine = false
 		}
 	}
@@ -106,10 +106,10 @@ func startsWithHashtag(src string) bool {
 	return false
 }
 
-func isPreprocessorLineMarker(src string) bool {
+func isLineMarker(src string) bool {
 	trimmedStr := strings.TrimSpace(src)
 	splittedStr := strings.Split(trimmedStr, " ")
-	if len(splittedStr) > 2 && splittedStr[0] == "#" {
+	if len(splittedStr) > 2 && splittedStr[0][0] == '#' {
 		_, err := strconv.Atoi(splittedStr[1])
 		if err == nil {
 			return true
--- a/src/arduino.cc/builder/ctags_target_file_saver.go
+++ b/src/arduino.cc/builder/ctags_target_file_saver.go
@@ -34,6 +34,8 @@ import (
 	"arduino.cc/builder/types"
 	"arduino.cc/builder/utils"
 	"path/filepath"
+	"strconv"
+	"strings"
 )
 
 type CTagsTargetFileSaver struct {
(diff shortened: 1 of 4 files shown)


src/arduino.cc/builder/ctags_target_file_saver.go
```diff
@@ -59,3 +67,53 @@ func (s *CTagsTargetFileSaver) Run(context map[string]interface{}) error {
...
+	lines := strings.Split(src, "\n")
+
+	saveLine := false
+	minimizedString := ""
+
+	for _, line := range lines {
+		if (saveLine || startsWithHashtag(line)) {
+			minimizedString += line + "\n"
+		}
+		if (containsAny(line, tofind)) {
```
matthijskooijman: Why no `isLineDirective()` check here? Sounds like it could somehow trigger unintended matches when the sketch location somehow occurs in the included code somewhere?

src/arduino.cc/builder/ctags_target_file_saver.go
```diff
@@ -59,3 +67,53 @@ func (s *CTagsTargetFileSaver) Run(context map[string]interface{}) error {
...
+
+func startsWithHashtag(src string) bool {
+	trimmedStr := strings.TrimSpace(src)
+	if (len(trimmedStr) > 0 && trimmedStr[0] == '#') {
+		return true
+	}
+	return false
+}
+
+func isLineDirective(src string) bool {
```
matthijskooijman: It seems this function is slightly mis-named. I would expect it checks for [#line directives](https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html), but it seems it actually checks for ["line markers" generated by the preprocessor](https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html). I guess changing the name should be changed correspondingly?

Also, it seems that any non-printing characters in the filenames will be escaped using \xxx escapes by the preprocessor. The current substring-based matching above doesn't handle these properly. I think the correct way to handle this, is to make `isLineDirective` (which should probably be renamed to `filenameFromLineMarker` or something?) actually parse the line marker, apply unescaping to the filename and return that?

src/arduino.cc/builder/ctags_target_file_saver.go
```diff
@@ -59,3 +67,53 @@ func (s *CTagsTargetFileSaver) Run(context map[string]interface{}) error {
...
 }
+
+func saveLinesContainingDirectivesAndSketch(src string, tofind []string) string {
+	lines := strings.Split(src, "\n")
+
+	saveLine := false
+	minimizedString := ""
+
+	for _, line := range lines {
+		if (saveLine || startsWithHashtag(line)) {
```
matthijskooijman: Why save all the # lines? AFAICS there are only line markers left in the output, and only the ones that are seen while `saveLine` is (or becomes) true are relevant I think?

matthijskooijman:
I finally got around to reviewing this. I think the idea is great and it should indeed solve all kinds of ctags problems, but I have some doubts about the current implementation (as indicated in the previous comments).

facchinm:
Hi Matthjis,
thanks A LOT for taking the time to review this! The implementation is clearly suboptimal and a lot of corner cases are not correctly handled but I'm happy you appreciate the idea. 
"Fake" line directives are saved only to keep things simple (like "save everything which starts with `#`" ) and saving them bring no side effect. 
The functions names need a fix and we really need to write tests for strange path cases, I'll take care of these issues soon :wink:

matthijskooijman:
I separated the fix and testcase into #154 (I merged them into a single commit) and added a bit more verbose commit message.

matthijskooijman:
I've improved (pretty much redid) the main subject of this PR in #156, so I'm closing this one in favor of that one.

PR #134, opened 2016-04-18
Reduce boilerplate code by rationalization of context data into a struct

cmaglie:
This pull-request aims to dramatically increase readability of the arduino-builder source code by removing a lot of boilerplate code. This change set involves a large part of the builder.

The rationale behind this PR has been partially discussed here: https://github.com/arduino/arduino-builder/pull/87#commitcomment-15187887 

Basically I'm trying to use an explicit golang `struct` instead of a map-of-strings-to-an-empty-interface `map[string]interface{}`. Doing so brings the following advantages

* We don't need context constants to enforce data consistency and we can get rid of all the `constants.CTX_....` because the compiler already checks correctness of the field name, for example:
```diff
-	context[constants.CTX_BUILD_PROPERTIES] = newBuildProperties
+	ctx.BuildProperties = newBuildProperties
```

* No need for "Unboxing" values of the map. The struct has fixed types so there is no need to cast the generic `interface{}` value to the correct type as we did before:
```diff
-func (s *AddMissingBuildPropertiesFromParentPlatformTxtFiles) Run(context map[string]interface{}) error {
-	packages := context[constants.CTX_HARDWARE].(*types.Packages)
-	targetPackage := context[constants.CTX_TARGET_PACKAGE].(*types.Package)
-	buildProperties := context[constants.CTX_BUILD_PROPERTIES].(props.PropertiesMap)
+func (s *AddMissingBuildPropertiesFromParentPlatformTxtFiles) Run(ctx *types.Context) error {
+	packages := ctx.Hardware
+	targetPackage := ctx.TargetPackage
+	buildProperties := ctx.BuildProperties
```

* Remove initialization code by exploiting golang default value. For example, since golang auto-initializes `bool` members to `false`, we could remove this initialization snippet:
```diff
-	if !utils.MapHas(context, constants.CTX_VERBOSE) {
-		context[constants.CTX_VERBOSE] = false
-	}
```

* The same optimization could be performed for arrays and maps because both are initialized to the empty set, so there is no need to check for the "existence" in advance:
```diff
-	if !utils.MapHas(context, constants.CTX_INCLUDES) {
-		context[constants.CTX_INCLUDES] = includes
-		return nil
-	}
-
-	context[constants.CTX_INCLUDES] = utils.AddStringsToStringsSet(context[constants.CTX_INCLUDES].([]string), includes)
+	ctx.Includes = utils.AppendIfNotPresent(ctx.Includes, includes...)
```


cmaglie:
@ArduinoBot build this please

src/arduino.cc/arduino-builder/main.go
```diff
@@ -249,8 +248,7 @@ func main() {
 		_, err := os.Stat(buildPath)
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
-			defer os.Exit(1)
-			return
+			os.Exit(1)
```
matthijskooijman: Isn't the defer useful here? IIRC deferring the exit was useful to make sure other defer and error handlers run, right? Or does that not apply here?

cmaglie: Oh good catch, that's a nice one that I forget to comment on the PR.
I kept this change in a single commit so it can be easily reverted in case.

> IIRC deferring the exit was useful to make sure other defer and error handlers run

what you said applies for deferred actions created in the current scope (i.e. the current running function and its callers), but since the `main` function, basically, does only command line parsing all the defers are instantiated inside "Commands" and when the control returns back to the main function all the internal defers are already out of scope and executed. Since the `main` function doesn't use `defer` I opted to skip the `defer os.Exit` construct.

The drawback is that new defers may be added in the `main` function in the future and it would be more future-proof to keep the `defer os.Exit` construct instead. Personally I prefer the "classic" behavior of the "exit" function, where the execution of the program is terminated immediately.

cmaglie: Another approach may be to make the `main` function just call sub-functions and convert the result into an exit code, something like:

```go
func main() {
    if err := parseCommandLine(); err != nil {
        exitWithError(err);
    }
    if err := runCommands(); err != nil {
        exitWithError(err);
    }
    os.Exit(0)
}

func exitWithError(err error) {
    var exitCode int
    // [...convert error into an exitCode...]
    os.Exit(exitCode)
}
```

matthijskooijman: I guess you could even use a more simple approach where the effective main function can just return an exit code, just like we know from C:

```
func main() {
    os.Exit(realMain())
}
```

And then the `realMain()` function just does what it needs and returns an exit code?

src/arduino.cc/builder/i18n/errors.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+	return errors.Errorf(Format(format, a...))
+}
+
+func WrapError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return errors.Wrap(err, 0)
+}
```
matthijskooijman: How is this i18n-related?

src/arduino.cc/arduino-builder/main.go
```diff
@@ -400,9 +387,11 @@ func printError(err error, printStackTrace bool) {
 func printCompleteError(err error) {
 	err = i18n.WrapError(err)
 	fmt.Fprintln(os.Stderr, err.(*errors.Error).ErrorStack())
+	os.Exit(1)
```
matthijskooijman: This seems unexpected to me - I wouldn't expect a function called "printCompleteError" to exit the current process. I guess the name should be changed? Or can this exit be removed when implementing the exit-related changes discussed elsewhere?

matthijskooijman:
I had a more thorough look over this PR, and overall it looks good. What a gargantuan amount of work, though, kudos for seeing it through :-)

However, I can't help but think we should be reducing the context object at some point, since effectively we're now just using global variables (hidden in a context struct) to pass function arguments and results around (very much commodore basic-style). We will need some other way to handle passing data around then, of course. Anyway, something to think about later, I think.

src/arduino.cc/arduino-builder/main.go
cmaglie: Yup that's another way too, BTW my preference is to handle everything with golang's `error` until the very end and use the `main` function to convert it into an error code. BTW I guess we can work this out after merging this PR :-)

src/arduino.cc/arduino-builder/main.go
```diff
@@ -170,85 +172,77 @@ func main() {
...
-	err, printStackTrace = setContextSliceKeyOrLoadItFromOptions(context, customBuildPropertiesFlag, buildOptions, constants.CTX_CUSTOM_BUILD_PROPERTIES, FLAG_PREFS, false)
-	if err != nil {
-		printError(err, printStackTrace)
-		defer os.Exit(1)
-		return
+	// FLAG_LIBRARIES
+	if librariesFolders, err := toSliceOfUnquoted(librariesFoldersFlag); err != nil {
+		printCompleteError(err)
+	} else if len(librariesFolders) > 0 {
+		ctx.LibrariesFolders = librariesFolders
```
matthijskooijman: There is a bug here, this should be `ctx.OtherLibrariesFolder` instead. `ctx.LibrariesFolders` is used by `LoadLibraries` to store the final complete and sorted list of library folders.

cmaglie: You're right, great catch! I'm fixing it...

src/arduino.cc/builder/i18n/errors.go
cmaglie: Not really related, but I've been forced to do this otherwise there is a cyclic import. The problem is in the `ErrorfWithLogger` method that wants a `Logger`, but the `Logger` interface is defined in `i18n` package, so the cyclical dependency.

For the moment I placed everything into the `i18n` package that is the solution with the minimum amount of changes. We can find a better way after merging the PR.

facchinm:
This PR works like a charm, and now it's really a pleasure to write new code without getting mad :smile: 
I'm for merging right now!

PR #135, opened 2016-04-20
Wipe anyway if the sketch name has changed

facchinm:
solves https://github.com/arduino/Arduino/issues/4862

If `build.path` is hardcoded in `preferences.txt` the current routine doesn't wipe the build directory, so we end up with a couple of `.o` defining `setup()` and `loop()`.
If we switch sketch, wipe anyway (note that, by default, every IDE window defines its own `build.path` so we don't wipe unless the target board has changed)

### DIFF
--- a/src/arduino.cc/builder/wipeout_build_path_if_build_options_changed.go
+++ b/src/arduino.cc/builder/wipeout_build_path_if_build_options_changed.go
@@ -34,9 +34,9 @@ import (
 	"arduino.cc/builder/gohasissues"
 	"arduino.cc/builder/i18n"
 	"arduino.cc/builder/types"
+	"encoding/json"
 	"os"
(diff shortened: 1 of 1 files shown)


matthijskooijman:
@facchinm, I'm working on some changes to support better debugging symbols, which indirectly also fix this bug (by implementing part of #86). I should have a PR somewhere this week, so you might want to hold off on merging this PR for a bit more.

facchinm:
Superseded by #144

cmaglie:
There is a problem in the loop logic: if the sketch has the same name the other build options are ignored. 

@matthijskooijman I've made #144 that fixes this and another problem. Since the current hourly builds are affected I'm going to merge #144 quickly and release upstream.

PR #142, opened 2016-05-04
Dependency parsing: if a file is not found trigger a full rebuild

### LINKED ISSUE #136: Dependency file parsing assumes one file per line
Apparently, the parsing of .d files assumes there is a single file on each line of the file. However, in some cases, presumably when filenames are short, there might be more than one in a single line. Consider this example from [this report](https://github.com/arduino/Arduino/issues/4880#issuecomment-211334109).

	C:\Users\Graham\AppData\Local\Temp\builde33ce9ddee6346054afe0349d71c85f0.tmp\libraries\UTFT\UTFT.cpp.o: \
	 F:\Arduino\libraries\UTFT\UTFT.cpp F:\Arduino\libraries\UTFT\UTFT.h \
	 C:\Users\Graham\AppData\Local\Arduino15\packages\arduino\hardware\sam\1.6.7\cores\arduino/Arduino.h \
	 (rest of the file removed)

The relevant source is here: https://github.com/arduino/arduino-builder/blob/master/src/arduino.cc/builder/builder_utils/utils.go#L222-L233

I guess the solution here is to do more proper parsing of the file, such as removing any escaped newlines, and then splitting the result into filenames (taking into account escaped spaces). I'm not entirely sure how reliable this can be done, since GNU make is known to be bad in handling spaces in filenames (though I think it works ok as long as no variables are involved).

cmaglie:
The missing file is probably due to a parsing error, better to trigger a rebuild in this case instead of exiting with an error. This is a quick workaround, the full solution is to properly parse .d files.

Fix #136

### DIFF
--- a/src/arduino.cc/builder/builder_utils/utils.go
+++ b/src/arduino.cc/builder/builder_utils/utils.go
@@ -222,7 +222,9 @@ func ObjFileIsUpToDate(sourceFile, objectFile, dependencyFile string) (bool, err
 	for _, row := range rows {
 		depStat, err := os.Stat(row)
 		if err != nil && !os.IsNotExist(err) {
-			return false, i18n.WrapError(err)
+			// There is probably a parsing error of the dep file
+			// Ignore the error and trigger a full rebuild anyway
+			return false, nil
 		}
 		if os.IsNotExist(err) {
 			return false, nil


cmaglie:
@ghlawrence2000 may you try this version of `arduino-builder` following the instructions above from ArduinoBot?

ghlawrence2000:
@cmaglie GREAT!!! That works perfectly, thank you!!

PR #144, opened 2016-05-06
Do not full-rebuild if only sketch path changed.

cmaglie:
This commit fix two problems:

1) the sketch path is used as-is in a regexp: this causes troubles if the path contains regexp control chars like `[-]().\` etc.
2) the sketch is not fully rebuild if the only thing that changes is the sketch path.

this PR supersedes #135


facchinm:
Wonderful

PR #145, opened 2016-05-06
Minor testing improvements

matthijskooijman:
This makes it a bit easier to share downloaded files between multiple working trees, and improves the document for running tests a bit.


README.md
```diff
@@ -76,14 +76,36 @@ go build arduino.cc/arduino-builder
...
+
+```
+go test arduino.cc/builder/test -run 'TestBuilderEmptySketch'
+go test arduino.cc/builder/test -run 'TestPrototypesAdder.*'
 ```
 
 In jenkins, use
 ```
-go test -timeout 60m -v ./src/arduino.cc/builder/test/... | bin/go-junit-report > report.xml
+go test -v arduino.cc/builder/test | bin/go-junit-report > report.xml
```
cmaglie: Well this timeout has been added not much for the initial download of tools but mainly for the ARM release: running the test natively on a Raspberry PI2 takes about 30 minutes. So probably it's best to keep it as default?

matthijskooijman: Right. Still, I'd like to keep the default commandline as short as possible, so that people can actually remember it and type it quickly, without having to copy-paste from the README every time. In any case, I'll add a note to the README about the ARM build.

matthijskooijman:
I added a note about ARM slowness to the README.

matthijskooijman:
Rebased on top of master, let's see if that fixes the ArduinoBot build.

PR #155, opened 2016-05-19
Improve handling of special characters in filenames

matthijskooijman:
This lets arduino-builder output correct #line directives, even if there are backslashes or doublequotes in the filenames. Backslashes were already handled in most cases, but this makes sure doublequotes are also handled. However, due to limitations in ctags, this does not actually make doublequotes in filenames work completely (prototypes are not generated), at least the build does not completely break now when there is a doublequote in a filename.


src/arduino.cc/builder/ctags/ctags_parser.go
```diff
@@ -229,7 +229,14 @@ func parseTag(row string) *types.CTag {
...
 	tag.FunctionName = parts[0]
-	tag.Filename = parts[1]
+	// This unescapes any backslashes in the filename. These
+	// filenames that ctags outputs originate from the line markers
+	// in the source, as generated by gcc. gcc escapes both
+	// backslashes and double quotes, but ctags ignores any escaping
+	// and just cuts off the filename at the first double quote it
+	// sees. This means any backslashes are still escaped, and need
+	// to be unescape, and any quotes will just break the build.
+	tag.Filename = strings.Replace(parts[1], "\\\\", "\\", -1)
```
cmaglie: Can we use `utils.ParseCppString`?

Having a special handling here looks weird, I'd use the full parsing even if ctags doesn't handle `\"`. Or maybe there are corner cases I'm missing?

matthijskooijman: I considered this, but:
 - `parseCppString` requires double quotes around the string, which should then be added here before passing the string to `parseCppString` (not really a problem)
 - If a double quote is present in a filename, it is escaped by gcc, but then ctags chops the string at the escaped double quote, causing the filename to end in a backslash. Parsing this by `parseCppString` will fail, so this would need extra special casing.

Mostly regarding the latter, it seemed better to just handle backslashes here, in a way that also "fixes" any double quotes broken by ctags.

cmaglie: > If a double quote is present in a filename, it is escaped by gcc, but then ctags chops the string at the escaped double quote, causing the filename to end in a backslash. Parsing this by parseCppString will fail, so this would need extra special casing.

I see, probably we can just ignore `parseCppString` errors, and just use the filename as is (`parseCppString` already return the input filename unchanged if the parsing fails).
In any case a file containing a double quote will not work, but having the full parsing in place may allows future versions of ctags to run properly (without the need to touch this part again).

matthijskooijman: This is true, yes. However, if ctags would properly handle escaping, it would need to unescape quotes and backslashes when reading the line directives, but it would need to escape backslashes and *tabs* when generating its output, since its fields are currently tab-separated. Alternatively, it would need to add quotes around the filename and escape quotes instead of tabs, meaning that tabs inside quotes are implicitely "escaped" (i.e. are not seen as a field boundary). In either case, arduino-builder needs more changes than you're proposing: splitting the ctags output on tabs will no longer work, since escaped tabs inside the filename should be skipped.

One alternative that would work without major changes with your suggestion is that tabs (and possibly other characters as well) are escaped using octal escapes (e.g. `\011` for tab). This is technically supported in a preprocessor string, though gcc line markers do not currently emit them, so `utils.parseCppString` does not currently parse them either.

Hence I'm inclined to leave this as it is now, and only improve the parsing here if it ever becomes relevant due to ctags improvements (which might not ever happen if we replace ctags soon).

PR #156, opened 2016-05-19
Only pass sketch sources to ctags 

matthijskooijman:
Previously, the .ino files from the sketch were concatenated,
preprocessed and then the entire preprocessing result was passed through
ctags. Since ctags isn't a perfect C/C++ parser, this would cause it to
sometimes break on some complex constructs from library or system header
files, preventing the entire sketch from compiling.

This commit filters the result of preprocessing to only include lines
that originate from the original sketch files and filter out any lines
from included system or library header files. This filtering happens
based on the line markers that the gcc preprocessor emits.

Because of this, the CollectCTagsFromSketchFiles pass can be removed.
This pass used to apply the same filtering, but on the tags generated by
ctags. Since now only sketch lines are fed into ctags, the resulting
ctags will obviously all originate from sketch files, so no need to
further filter them.

With this change, all sketches from the test suite still compile as
expected. However, there is a fair chance that with this change, ctags
will fail to parse some sketches. In general, to parse a C or C++ file,
you need to keep track of the type environment and symbol table, since
identifiers can either be variables or types, depending on the context.
When lines from included files are cut off, this information is not
available, and ctags will have to guess (though it might be doing that
anyway, not sure if it normally does completely proper parsing).

This PR achieves the same as #131, but has a more robust implmentation. Only the last commit really belongs to this PR, the first few commits come from #150 and #155. This PR should be merged carefully, as it might break sketches. It is probably best to merge this after 1.6.10, but be prepared to revert it turns out to cause regressions.


igrr:
I have compiled all the test cases in ESP8266 Arduino core project with arduino-builder from this pull request and can confirm that nothing got broken.

This pull request also fixes the long-standing issue with the following sketch, both for SAMD core and ESP8266 core:
```c++
#undef min
#undef max
#include <vector>
#include <memory>
#include <functional>
void setup() {
  test();
}
void loop() {}
void test() {}
```
Previously this did not compile because ctags failed to parse something inside STL headers.

So huge :+1: from me.

PR #157, opened 2016-05-30
It only builds with go >= 1.6.2

bogdanr:
Changed the readme so it reflects that it only builds with go greater than 1.6.2


cmaglie:
I've just tried to build it with go 1.5.4, from a clean workspace, and everything went smooth, which errors are you getting?

PR #158, opened 2016-06-01
Fix compilation of a library that contains a "utility" file

matthijskooijman:
Libraries can contain a "utility" directory, which is then added to the
include path and whose contents are compiled. However, the code that
handled this only checked for existence of the "utility" path, not if it
was actually a directory. If a file with that name existed, it would be
treated as a directory, breaking the build. This happens for example
with the StandardCplusplus library.


facchinm:
LGTM

PR #165, opened 2016-07-04
Add specifier to generated tool property keys

facchinm:
port of https://github.com/arduino/Arduino/pull/4795 to golang
should solve https://github.com/arduino/Arduino/issues/5042

This is far less powerful than Java version since the tools are not tangled with json package_index but all the information are extracted from the folder name.
Since the folder name only has a "vendor" subpath we cannot correlate the tools with the required version but only try to match the architecture being compiled with the tool vendor.


agdl:
:+1: Working with red bear labs hardware installed!

facchinm:
This PR breaks 4 tests (TestSetupBuildProperties, TestSetupBuildPropertiesWithMissingPropsFromParentPlatformTxtFiles, TestLoadTools, TestLoadLotsOfTools).
They are not really "broken" but the folder structure of `downloaded_tools` doesn't match board manager's so the builtin tools are now "visible", while they were hidden by `downloaded_tools` before.
This could be a problem when updating AVR core with new tools (https://github.com/arduino/Arduino/pull/5021) on older IDEs 

@matthijskooijman @cmaglie , any idea on how we can solve this in a sane way (without parsing package_index json, which would be the best option in the long term)?

facchinm:
Supoerseded by https://github.com/arduino/Arduino/issues/5199

PR #173, opened 2016-08-01
Allow the builder to be aware of package_index properties

facchinm:
AVR core 1.6.12 introduced a new toolchain but the mighty builder still had no clue about it. There is no magic in there, simply some key->value substitution which accidentally lead to a massive breakage, both for arduino avr core and for 3rd party cores.

To solve this issue in a sane way (but without providing a full package manager yet) this PR introduces a library (`json_package_index`) which parses the provided package_index folder, extracts the useful information from there and uses these information to avoid breaking the build.

Passing `-package-index` flag with a folder will allow the builder to know which version is needed by any core. 

I'll update this PR as soon as we'll start testing it massively (a link to the IDE PR builds will come soon)


facchinm:
IDE builds will appear here https://github.com/arduino/Arduino/pull/5197

facchinm:
Putting this PR in standby until we understand the future architecture of IDE related tools

facchinm:
Superseded by arduino-cli

PR #175, opened 2016-08-02
Include detection caching

matthijskooijman:
This PR adds caching to the include detection mechanism, greatly improving the speed of builds with a lot of files.

This code was only just finished and has not seen a lot of testing yet. Unit tests should be added for it as well. However, I wanted to get the code out for review ASAP to collect some comments.

This PR is made on top of #174, so the first dozen or so commits should be ignored until that is merged.


PaulStoffregen:
I'm trying this now with the Teensy Audio library.  Seems to make no difference.  The "g++ -E" step is being run on all the (many) library files every time I click Verify.

PaulStoffregen:
It does make a huge speedup when I compile the Ethernet library AdvancedChatServer example for Arduino Mega.

PaulStoffregen:
Hmmm... it works if I try IRremote's IRrecvDumpV2 example, compiling for Teensy 3.2.  So it does seem to work, but there's something about the Teensy Audio library that's causing it to not use the cache data at all.

I'm testing on Linux 64 bit.

To reproduce this, start with a clean copy of 1.6.10 and then run [this installer](https://forum.pjrc.com/threads/35700-Teensyduino-1-30-Beta-2-Available) to all the Teensy stuff, then replace arduino-builder.  Select Teensy 3.2 from Tools > Boards, and then open File > Examples > Audio > Synthesis > Guitar.

PaulStoffregen:
When it does work, wow, really speeds things up!  Hopefully this can work for the audio lib... it's a very large library with a *lot* of files, so perhaps a tough test case?

matthijskooijman:
@PaulStoffregen, thanks for testing and providing a broken testcase, I'll dig into it (but probably not before this weekend or next week).

You link to an installer for Teensyduino, but is there also another installation method available? e.g. just a zip to unpack, or perhaps a boards manager url? I don't really like to run arbitrary binary installers without knowing what they'll do exactly (though it's not a blocking problem, just wondering). Is there anything in teensyduino that cannot be expressed in a board manager json file?

I just ran the installer and tested the example you mention, and I can reproduce the problem.

matthijskooijman:
I found a bit of time in the train and couldn't resist looking into this already :-) Turns out this caching code exposes two existing bugs in the code, which were conveniently hiding each other:
 1. When compiling a library, its `utility` directory is added to the include path if it exists (only of the library itself, not of other libraries). However, when doing include detection for a library, the `utility` directory is *not* put on the include path. In case of the Teensy Audio library this makes (for example) "sqrt_integer.h" show up as a missing include file, but arduino-builder cannot find any library for it, so it is intended to bail out (but see next).
 2. When no library is found for an include file, arduino-builder is supposed to bail out. It does this by running the preprocessor again, but instead of parsing the error, the error is shown to the user. However, this preprocessor is run on the main sketch file, not on the file that was being processed. If that compiles succesfully, no error is returned so include processing continues with the next source file in the list.

The reason that the second bug doesn't normally cause any actual problems is that if an include is really missing, the error about it will show up later when actually compiling the problematic file.

The reason the first bug doesn't normally cause problems is that when such an include from the `utility` folder is not found, due to the second bug, this only aborts include detection for the current file, and continues with the rest of the files. When actually compiling the problematic file, the `utility` directory is added to the include path as expected, and the compile compiles succesfully.

The first bug could cause problems if a utility-path include is above an include for another library, and that other library is not included from other places (such as the sketch). This would cause inlude detection to miss that other library and the build will fail later on. I'll write a testcase to confirm this behaviour.

As for fixing these, I think the fixes are fairly simple:
 1. Add the `utility` folder to the include path for the current library. The `SourceFile` class I introduced give the include detection sufficient context to know a source file is from a library and to add the utility path to the include path.
 2. An obvious fix would be to run the preprocessor on the current file, but I think just printing the error output from the previous preprocessor run (which was saved in order to parse it for #include error messages) is even more efficient and true (less likely to suddenly not fail due to subtle differences in the compiler commandline).

Fixing this will probably take a while longer, my train ride is just over :-p

PaulStoffregen:
Oh, I didn't realize the utility folder would matter.

I'm continuing to use this as I work on another project.  This particular code has a lot of extra .cpp and .c and .h files in the sketch folder.  It seems to be caching those very nicely.  Saves a lot of time when working with more complex code using a lot of files!  Very nice.

Looks like there may be a similar issue with my heavily modified copy of the SD library.  Can't reproduce it with the built-in SD lib.  If you really want another test case, I could trim this down to something I can share.

matthijskooijman:
> Looks like there may be a similar issue with my heavily modified copy of the SD library.

Does it have a utility folder? If so, it's likely the same issue.

PaulStoffregen:
Both have utility folders, but one is src/utility, the other is just the utility without any src folder.

matthijskooijman:
Hm, src/utility should not be affected, since that isn't added to the include path during compilation. If you could provide a testcase, that would be useful (I'm also seeing if I can fix the other issue now, so perhaps you could await that before spending time on it).

PaulStoffregen:
Yup, the one with src/utility seems to always work.

Even with minor issues, excellent work on this caching.  It's going to make a lot of Arduino users much happier.

matthijskooijman:
I just pushed a fix for the previously discovered issues. With this, caching works as expected for the Audio library example you indicated. Let me know if you find other issues, thanks again for testing!

FrankBoesing:
Works great ! 
esp. for my very large projects (with large libs like mp3 and so on, for Teensy) , it is a valuable help.

matthijskooijman:
I've found one bug/limitation with this implementation: When compilation does not complete successfully, a lot of includes are also re-detected even though nothing changed for them. This is because the caching relies on .d files generated by compilation, so if the compilation finds an error and aborts, the caching works only for file up to and excluding the file with the error.

The fix for this would be to let the include detection gcc command actually generate .d files as well, but that requires modifying the recipes, which might be somewhat tricky in terms of compatibility. I'm not sure if there is another easier fix for this...

cmaglie:
> I've found one bug/limitation with this implementation: When compilation does not complete successfully, a lot of includes are also re-detected even though nothing changed for them. This is because the caching relies on .d files generated by compilation, so if the compilation finds an error and aborts, the caching works only for file up to and excluding the file with the error.

In this case the cache is lost and the library detection is performed again from the beginning, but this is not worse compared to the current code when this happens every time, right?

If yes, I'm for merging anyway since it's a very good improvement, even with this limitation.

matthijskooijman:
> In this case the cache is lost and the library detection is performed again from the beginning, but this is not worse compared to the current code when this happens every time, right?

Correct (it is even still a little bit better, since the cache will still be used up to the file with the error).

PR #178, opened 2016-08-29
Fix UTF8 filenames not being matched by gcc -e output parsing

facchinm:
Starting from https://github.com/arduino/arduino-builder/commit/81eadf324ef2f3d14034452bd40bd2f2eb75ff98, `ctags` is fed only with sketch code after the `gcc -e` pass.
If the filename contains non-ASCII characters, the string matching at https://github.com/arduino/arduino-builder/blob/master/src/arduino.cc/builder/filter_sketch_source.go#L59 returns `false` because `utils.ParseCppString` only treats ASCII strings.

This patch fixes this behaviour and adds some tests to demonstrate the functionality. All tests are passing.


facchinm:
Superseded by #195

PR #189, opened 2016-11-11
Move sketch size calculation to Golang

facchinm:
ATM Java IDE takes care of calculating the size of a sketch (https://github.com/arduino/Arduino/blob/ec2e9a642a085b32701cf81297ee7c1570177195/arduino-core/src/processing/app/debug/Sizer.java)
This function can be safely moved to the builder BUT there are a couple of gotchas:
- The i18n layer only accepts strings, so the numbers are converted to strings before "logging" them. This result in "thousands" not being separated by commas or dots (based on the locale)
~~- The error path is "concurrent" (Java side) with the stdout path, so the output is perfect when run in go but becomes confused when displayed through "MessageSiphon". The 200ms delay before returning an error solves this (in a very ugly way)~~

The commits should be squashed and the constants moved to constants.go

@cmaglie @matteosuppo @00alis @mastrolinux


matteosuppo:
I'm getting

```Sketch uses 3624 bytes (12%!)(MISSING) of program storage space. Maximum is 28672 bytes.

Global variables use 148 bytes (5%!)(MISSING) of dynamic memory, leaving 2412 bytes for local variables. Maximum is 2560 bytes.``` 

Is it a locale issue?

**Comment by facchinm** [comment]
@matteosuppo  Could be... Which locale are you using? Mine is `en_GB.UTF-8`

**Comment by matteosuppo** [comment]
```admin@api-01-dev:~/builder$ locale
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE=en_EN.UTF-8
LC_NUMERIC=it_IT.UTF-8
LC_TIME=it_IT.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=it_IT.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=it_IT.UTF-8
LC_NAME=it_IT.UTF-8
LC_ADDRESS=it_IT.UTF-8
LC_TELEPHONE=it_IT.UTF-8
LC_MEASUREMENT=it_IT.UTF-8
LC_IDENTIFICATION=it_IT.UTF-8
LC_ALL=
```

matteosuppo:
Works for me

src/arduino.cc/builder/phases/sizer.go
```diff
@@ -0,0 +1,160 @@
...
+	out, err := builder_utils.ExecRecipe(properties, constants.RECIPE_SIZE_PATTERN, false, false, false, logger)
+
+	if err != nil {
+		return errors.New("Couldn't determine program size: {0}")
+	}
+
+	// force multiline match prepending "(?m)" to the actual regexp
+
+	if len(properties["recipe.size.regex"]) > 0 {
+		textRegexp := regexp.MustCompile("(?m)" + properties["recipe.size.regex"])
```
cmaglie: `regexp.Compile` should be used instead of `regexp.MustCompile` because is not said that `recipe.size.regex` contains a valid regular expression. In case of error we should print a warning and ignore the regexp as it was missing.

src/arduino.cc/builder/phases/sizer.go
```diff
@@ -0,0 +1,160 @@
...
+
+	properties := buildProperties.Clone()
+	properties[constants.BUILD_PROPERTIES_COMPILER_WARNING_FLAGS] = properties[constants.BUILD_PROPERTIES_COMPILER_WARNING_FLAGS+"."+warningsLevel]
+
+	maxTextSizeString := properties["upload.maximum_size"]
+	maxDataSizeString := properties["upload.maximum_data_size"]
+
+	if maxTextSizeString == "" {
+		return nil
+	}
```
matthijskooijman: Why this limitation? Even for boards without a max size, showing the current size is useful (and the IDE does it currently).

src/arduino.cc/builder/phases/sizer.go
```diff
@@ -0,0 +1,160 @@
...
+
+	logLevel := constants.LOG_LEVEL_INFO
+
+	maxTextSize, _ := strconv.Atoi(maxTextSizeString)
+	maxDataSize := -1
+
+	if maxDataSizeString != "" {
+		maxDataSize, _ = strconv.Atoi(maxDataSizeString)
```
matthijskooijman: What does Atoi return on an error? Is the first return value even defined in that case? Or should the err value (that is now ignored) be checked instead? This also applies to some other Atoi calls.

src/arduino.cc/builder/phases/sizer.go
matthijskooijman: This also means the printing code above should be changed to not print sketch size when the regex is missing.

src/arduino.cc/builder/phases/sizer.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var textSize, dataSize, eepromSize int
+
+	err := execSizeReceipe(properties, logger, &textSize, &dataSize, &eepromSize)
+	if err != nil {
+		return err
+	}
+
+	logger.Fprintln(os.Stdout, logLevel, "Sketch uses {0} bytes ({2}%%) of program storage space. Maximum is {1} bytes.", strconv.Itoa(textSize), strconv.Itoa(maxTextSize), strconv.Itoa(textSize*100/maxTextSize))
+	if dataSize > 0 {
```
matthijskooijman: This hides the message when dataSize is 0, though that can still be a valid size. Perhaps the "error" value should be < 0?

src/arduino.cc/builder/phases/sizer.go
facchinm: I agree, but the original code returns if no max size is specified https://github.com/arduino/Arduino/blob/21ff728c59c1a2fb138348f4e1f6cb4999ff27f5/arduino-core/src/cc/arduino/Compiler.java#L303 .
In fact, max sketch size is compulsory right now :smile:

src/arduino.cc/builder/phases/sizer.go
matthijskooijman: What do you mean by "report an error"? Panic?

Given that [Atoi returns (int, err)](https://golang.org/pkg/strconv/#Atoi), I would expect it will return a non-nil error value, but I'm wondering what the returned int value is in that case.

src/arduino.cc/builder/phases/sizer.go
matthijskooijman: Ah, missed that part. I probably remembered the data max size being optional, no the text size. Still, it might be nice to make the max text size optional now, while we're here?

matthijskooijman:
Looks good. As for "to be squashed" commits, do you know about `git commit --fixup` (combined with `git rebase --interactive --autosquash`) that can somewhat automate this?

facchinm:
Sure, I only wanted to keep track of the changes to avoid overwriting your comments :smile: 
If everything is ok (and approved by @cmaglie ) I'll fixup the commits before committing

matthijskooijman:
My suggestion would be to use `--fixup`, which creates a new commit with a special commit message that you could have pushed now, and then use the `--autosquash` later to squash to commit in the right place (without having to think about where that is). Not really relevant here, since I think you'll squash them into a single commit, but it can be useful for bigger PRs.

cmaglie:
I've added some more errors check and a test.

PR #191, opened 2016-11-18
Fix prototype callback generation if direct reference is used

### LINKED ISSUE #180: Auto generated function declarations no longer works for callbacks assigned at variable declaration
Here's and example that will break the auto generation of function declarations in the IDE, not sure for how long but I've gotten regular support request for a good while now.
The only thing I know, this used to work.

```
// say you have a library that accepts a callback
class BUG {public:BUG(void (*fn)(void)){}};

//void makeItBreak(){} //uncomment to fix
BUG b(makeItBreak); //this will break

void setup(){}
void loop(){}
void makeItBreak(){}
```

facchinm:
Fixes  arduino/arduino-builder#180
Should be refactored when merging   arduino/arduino-builder#190

### DIFF
--- a/src/arduino.cc/builder/ctags/ctags_parser.go
+++ b/src/arduino.cc/builder/ctags/ctags_parser.go
@@ -236,7 +236,6 @@ func parseTag(row string) *types.CTag {
 
 	parts = parts[2:]
 
-	signature := ""
 	returntype := ""
 	for _, part := range parts {
 		if strings.Contains(part, ":") {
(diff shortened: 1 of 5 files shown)


matthijskooijman:
From the commit messages and testcase it is not entirely clear to me what kind of code this is actually catering for. Could you perhaps expand on that (and probably update the commit message as well for future reference)?

facchinm:
This is actually a fix for the forward declaration generator; when guessing where to insert the declarations, it searches for function pointers in the form `&function`, while the preferred Arduino style for FP is `function` as in `attachInterrupt`.
With this PR it also searches for `function` only if the function signature is `()` (to avoid too many false positives, since it's only searching for a string ignoring the semantics)

Recalling https://github.com/arduino/arduino-cli/issues/1944, we can't even diagnostic these kind of problems because the prototype is generated, but too late.

PR #192, opened 2016-11-18
Naive way to merge sketches and bootloaders in bin format

facchinm:
Works with SAMD platform (at least)


matthijskooijman:
This looks a bit like we're hardcoding too much. The current merging of .hex files is already a special case, and deciding what to do based on the extension of the generated file or bootloader file (which I think is what happens now?) might be too far down that rabbit hole? Would it not be better to let the platform.txt / boards.txt specify what kind of merging needs to happen and/or with what files? The actual merging methods can still be somewhat hardcoded, but at least they're more extensible then.

Also, this introduces the build.ldscript variable, but it seems unused by the merging code? What's up with that?

facchinm:
This was a very special purpose PR (thus its hackish feeling :stuck_out_tongue: )

`hex` format is position independent, so merging the bootloader after the sketch will never produce an unbootable file (at least the bootloader part will be flashed correctly)

The correct approach for merging `bins` would include a "partitions" list csv (as happens on ESP8266) to extract the correct locations; the ldscript way is transparent but too SAMD dependant (and it doesn't take into account an address space offset).

Issue #193, opened 2016-11-19
Rebuild of core.a, even when none of its .o files have changed


PaulStoffregen:
Long ago in Arduino 1.6.5, rebuilding core.a was skipped when all of its .o files were unchanged.  This used to print in the verbose info:

    Using previously compiled file: /tmp/build2611869353429022717.tmp/core.a

On Windows with slower disk drives, this made quite a noticeable speedup.  Hopefully the builder can implement this someday?

PaulStoffregen:
Here's the old Java code which used to perform this function:

https://github.com/arduino/Arduino/blob/1.6.5-r5/arduino-core/src/processing/app/debug/Compiler.java#L1064

cmaglie:
Fixed by #194

PR #194, opened 2016-11-21
Use previously compiled "core.a" if core files are unmodified

facchinm:
Solves #193

Needs to add -MMD to .S.flags targets on all cores (otherwise the core gets recompiled everytime)

Signed-off-by: Martino Facchin <m.facchin@arduino.cc>


matthijskooijman:
Looks good to me (not thoroughly reviewed or tested, though).

PaulStoffregen:
I'm using it now in a copy of 1.6.12.  Seems to work well.

Will do much with it over the next several days....

PR #195, opened 2016-11-23
Fix UTF8 filenames not being matched by gcc -e output parsing (round 2)

cmaglie:
Supersedes #178 

/cc @facchinm


matthijskooijman:
Looks good to me!

Issue #70, opened 2015-11-23
WARNING: Spurious .git folder in library


ffissore:
_From @hryamzik on November 23, 2015 1:18_

Due to #3237 there shouldn't be such a warning but I do get in in Arduino IDE v1.6.6.

_Copied from original issue: arduino/Arduino#4204_

NicoHood:
I still get this warning for other folders like this:
`WARNING: Spurious .dev folder in 'HID-Project' library`

It would be cool to remove this as well.

ffissore:
The silenced warnings were about SCCS sofware. It's expected that warnings about actual spurious directories are notified

ladyada:
heya @ffissore - could .github also be ignored? its used to store issue templates and (likely in the future) other github configurations and setups, and will get included in the library zip. 

thank you :)

ffissore:
Hi @ladyada sorry you missed the news but I don't work for arduino any more. See [this message](https://groups.google.com/a/arduino.cc/forum/#!msg/developers/YJLX6AZHem4/kpsLef7XBAAJ). My former colleagues @facchinm @cmaglie and @sandeepmistry will take care of your issue

ladyada:
@ffissore apologies for the mixup - and good luck with all your endeavours! :)

matthijskooijman:
@ladyada, I created a PR for that: https://github.com/arduino/arduino-builder/pull/162

Thanks for the suggestion :-)

ladyada:
love it, thanks!

PR #198, opened 2016-12-18
Don't forbid unknown files in a library

mfalkvidd:
Warning about unknown folders was considered bad in
https://github.com/arduino/Arduino/pull/1692
but the problem is still present in arduino-builder.

This has caused issues like
https://github.com/arduino/arduino-builder/issues/70#issuecomment-158994311
which resulted in
https://github.com/arduino/arduino-builder/pull/162

The original arguments in Arduino/pull/1692 still apply:
* It breaks forward compatibility if we later add more files or
  directories to the library format.
* It breaks for people who want to have some extra stuff in their
  library. We can't keep a list of "allowed" stuff, since there will always
  be stuff missing.

Signed-off-by: Mikael Falkvidd git@mjo.se <git@mjo.se>

### DIFF
--- a/src/arduino.cc/builder/constants/constants.go
+++ b/src/arduino.cc/builder/constants/constants.go
@@ -192,7 +192,6 @@ const MSG_USING_CACHED_INCLUDES = "Using cached library dependencies for file: {
 const MSG_WARNING_LIB_INVALID_CATEGORY = "WARNING: Category '{0}' in library {1} is not valid. Setting to '{2}'"
 const MSG_WARNING_PLATFORM_MISSING_VALUE = "Warning: platform.txt from core '{0}' misses property '{1}', using default value '{2}'. Consider upgrading this core."
 const MSG_WARNING_PLATFORM_OLD_VALUES = "Warning: platform.txt from core '{0}' contains deprecated {1}, automatically converted to {2}. Consider upgrading this core."
-const MSG_WARNING_SPURIOUS_FILE_IN_LIB = "WARNING: Spurious {0} folder in '{1}' library"
 const MSG_WRONG_PROPERTIES_FILE = "Property line '{0}' in file {1} is invalid"
 const MSG_WRONG_PROPERTIES = "Property line '{0}' is invalid"
 const PACKAGE_NAME = "name"
--- a/src/arduino.cc/builder/libraries_loader.go
+++ b/src/arduino.cc/builder/libraries_loader.go
@@ -153,16 +153,6 @@ func makeNewLibrary(libraryFolder string, debugLevel int, logger i18n.Logger) (*
 		return nil, i18n.WrapError(err)
 	}
 
-	if debugLevel >= 0 {
-		for _, subFolder := range subFolders {
-			if utils.IsSCCSOrHiddenFile(subFolder) {
-				if !utils.IsSCCSFile(subFolder) && utils.IsHiddenFile(subFolder) {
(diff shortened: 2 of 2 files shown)


facchinm:
Hi @mfalkvidd , the warning was reintroduced one commit later (https://github.com/arduino/Arduino/commit/3bec9121cd1ae442c87db673747ab2fef27d32ad) and it's quite useful to enforce libraries uniformity. The hidden VCS folders will evolve over time, but since it's a warning there will be no functional problem in compiling a new library with an old arduino-builder

fallberg:
Sorry, but this warning is just stupid. Why do you show a warning for something that had nothing to do with your IDE? Why can't people use well defined standard directory structures for organizing things just because Arduino IDE references files in the tree? All this warning does is giving concern and worry to inexperienced users.

tbowmo:
So we should just tell users, "Yes, you see a warning, but it's nothing".. Which is just plain wrong..

imho all warnings should be dealt with.. And we could just as well start by educating this to users.

Now this warning will make it impossible to make a build / linter tools directory and distribute via the repositories. This shouldn't be visible to normal users, as they shouldn't need to change anything in this directory, but still be available to advanced developers.

bblacey:
It is often considered best practice to name supporting files as 'dot' files so they are hidden during normal directory traversal yet available to tools within the repo namespace for things like Travis-CI that names its config file .travis.yml.  Furthermore, users often place their continuous-integration files in folders that begin with a dot such as .travis; the same goes for other hosted CI services like Circle-CI and the list goes on including [this recent example](https://github.com/mysensors/MySensors/tree/development/.mystools)

Bottom-line:  The approach used for the Arduino IDE is dated relative to the broader perspective of how to develop manage and test a product within a git repo that transcends multiple-tool usage.  Said another way, the IDE is fully within it's right to complain about unknown files within a sketch directory that is defined by the IDE but the IDE is just a **single** tool amongst many that is used within the scope of a repo and should not interfere with usage of other tools within the same namespace as it is currently doing.

Currently, the IDE attempts some tolerance by ignoring well-known dot directories such as .git, .svn, etc. but this is hardwired into the source code (not extensible by users)...  

One potential solution would be to use the well-established convention (many, many tools use this convention) of implementing an .arduinoconfig to provide the IDE with namespace-specific configuration settings. This would be a convenient place to provide IDE configurations for a given namespace with one configuration setting being a list of items within the namespace to ignore.

matthijskooijman:
@facchinm, I'd also consider reopening this issue and removing the warning, since it doesn't seem to add much. What are we trying to keep out of libraries? Object files? Should we perhaps add specific warnings instead?

facchinm:
I believe the warning is there only as a (gentle) reminder for library developers; someone could say it is wrong to ship a library the stray useless files in it (and versioning folders are useless from a user's POW). 
For example, if you download (not clone) a library from github it loses all the versioning info. This is also the standard behaviour of every FOSS library available.
Following this philosophy, whitelisting dot-folders is wrong, since these warnings should be reported anyway if the developer activates a flag, while they should be invisible by default. 

Could adding a particular tag to the warning and filtering it in IDE UI be a good solution for everyone?

fallberg:
@facchinm well you can definitely not make the judgment from Arduino IDE if files and folders in a library repository is useless or not. It could well be used in other non Arduino contexts where the files and folders are very much in use. So at the very least, there should be a possibility for a library to whitelist its own files.

PaulStoffregen:
When/if you add new features to the library spec (at least ones involving new folders & files), this will slow their adoption because library authors usually wish to support at least the last couple years of IDE releases which many people still have installed.  Maybe that's ok?

PR #205, opened 2017-01-19
Report the message levels on the human logger

matteosuppo:
Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This tool generates function prototypes and gathers library paths, providing `gc
 
 * `-core-api-version`: Optional, defaults to "10600". The version of the Arduino IDE which is using this tool.
 
-* `-logger`: Optional, can be "human" or "machine". Defaults to "human". If "machine", messages emitted will be in a format which the Arduino IDE understands and that it uses for I18N.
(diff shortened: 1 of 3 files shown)


matthijskooijman:
Code looks good. What's the intended usecase for this logger?

matteosuppo:
On create we use the human logger, but we would like to use different colors for the different levels.

This way we could also identify what's the issue in this case, for example:

human
```Sketch uses 904 bytes (2%) of program storage space. Maximum is 32256 bytes.
Global variables use 3009 bytes (146%) of dynamic memory, leaving -961 bytes for local variables. Maximum is 2048 bytes.
Not enough memory; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing your footprint.
```

humantags
```[info] Sketch uses 904 bytes (2%) of program storage space. Maximum is 32256 bytes.
[info] Global variables use 3009 bytes (146%) of dynamic memory, leaving -961 bytes for local variables. Maximum is 2048 bytes.
[error] Not enough memory; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing your footprint.```

PR #210, opened 2017-01-31
Relax library matching rules when detecting libraries

cmaglie:
This allows to set the correct library priority, for example, if the library is named "My Library" but the include file is "my_library.h".

/cc @matteosuppo @facchinm

### DIFF
--- a/src/arduino.cc/builder/resolve_library.go
+++ b/src/arduino.cc/builder/resolve_library.go
@@ -56,7 +56,6 @@ func ResolveLibrary(ctx *types.Context, header string) *types.Library {
 	}
 
 	if len(libraries) == 1 {
-		markImportedLibrary[libraries[0]] = true
 		return libraries[0]
 	}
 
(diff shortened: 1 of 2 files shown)


src/arduino.cc/builder/resolve_library.go
```diff
@@ -55,14 +55,14 @@ func ResolveLibrary(ctx *types.Context, header string) *types.Library {
 		return nil
 	}
 
-	if len(libraries) == 1 {
-		return libraries[0]
-	}
-
 	if markImportedLibraryContainsOneOfCandidates(markImportedLibrary, libraries) {
 		return nil
 	}
```
matthijskooijman: Note that this should actually never occur at all, if the header is available, it is never returned as missing by the preprocessor (this was different when include scanning was still regex-based instead of preprocessor-based). Keeping this `if` in might be a good idea, though, but perhaps it should be marked as "just in case" with a comment.

PR #213, opened 2017-02-09
Archive compiled core and use on subsequent compilations

facchinm:
Very WIP but almost working good


src/arduino.cc/builder/builder_utils/utils.go
```diff
@@ -258,6 +289,24 @@ func nonEmptyString(s string) bool {
 	return s != constants.EMPTY_STRING
 }
 
+func CheckIfRecompileIsAvoidable(corePath string, archiveFile string) bool {
```
facchinm: Correct, the timestamps are only extracted from `core` subfolder, while they should be applied to the whole folder. I'm pushing another commit to fix this

src/arduino.cc/builder/builder_utils/utils.go
```diff
@@ -366,3 +415,67 @@ func ExecRecipeCollectStdErr(buildProperties properties.Map, recipe string, remo
...
+	}()
+	if _, err = io.Copy(out, in); err != nil {
+		return
+	}
+	err = out.Sync()
+	return
+}
+
+func GetCoreArchivePath(fqbnToUnderscore string) string {
+	return os.TempDir() + "/core_" + fqbnToUnderscore + ".a"
```
facchinm: This is only a temporary way to store them :smile: A proper db (or folder structure) is surely the way to go

src/arduino.cc/builder/setup_build_properties.go
```diff
@@ -70,6 +70,7 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
 	buildProperties[constants.BUILD_PROPERTIES_RUNTIME_PLATFORM_PATH] = targetPlatform.Folder
 	buildProperties[constants.BUILD_PROPERTIES_RUNTIME_HARDWARE_PATH] = filepath.Join(targetPlatform.Folder, "..")
 	buildProperties[constants.BUILD_PROPERTIES_RUNTIME_IDE_VERSION] = ctx.ArduinoAPIVersion
+	buildProperties[constants.BUILD_PROPERTIES_FQBN] = strings.Replace(ctx.FQBN, ":", "_", -1)
```
facchinm: You are right, `buildProperties[constants.BUILD_PROPERTIES_FQBN]` should contain the actual fqbn, however I'm going to apply the substitution later to make sure the filename is valid on all OS

PaulStoffregen:
Is there an @ArduinoBot build?  I'd like to give this a try here, but I don't have the Go stuff set up.

On Teensy, I make much more use of defining things from boards.txt and the custom menus, which are then referenced by platform.txt.  The entire core library really needs to be rebuilt when those change.  I know most 3rd party boards just copy platform.txt with only minimal changes, so maybe this is a corner case... but I'm pretty concerned about core.a not rebuilding...

facchinm:
The bot is still in a coma, I'm parsing here the link for the artefacts 
http://downloads.arduino.cc/PR/arduino-builder/arduino-builder-213.zip

facchinm:
@PaulStoffregen if the custom menus selection changes the fqbn (like `mega:cpu=2560`) it won't use the archived core, otherwise it will (at the moment)

facchinm:
@ArduinoBot build this please.

src/arduino.cc/builder/builder_utils/utils.go
matthijskooijman: How does this interplay with referenced cores? E.g. when a platform has a platform.txt but does not contain the actual core, but instead references (e.g.) the arduino core? I believe compilation options could then come from either platform?

In general, I wonder if this can be more thoroughly covered? Ideally, a compiled core would keep a file (like the .d files for each .o file) that shows what files were used to create the .a file, so these could all be checked for modifications, as well as noting the compile options / build properties used to build the core. Even more general would be to log the complete compiler commands executed to build a .a file, and then compare that to the commands that *would be* executed on a subsequent run, and only use them if the commands are exactly equal? This is a bit similar to the caching for include detection that is now used.

facchinm: There are some tricky details, indeed, which should be tackled in a saner way. The inter-dependencies problem is one (and could be solved by tracking also the parent core). Note that the current builder does not wipe the environment if `platform.txt` changes (because it's not part of the .d dance). Tracking dependencies via .d files is really handy for libraries or mutable code, while the core is not really mutable (in Create it's even "static"). To be even more conservative, one could track all the modification dates on all core folders and wipe if even one file is newer than the precompiled core (this could be time consuming in case of slow disks or lots of cores).

Issue #218, opened 2017-03-06
Txt files parsing should not fail in case of inline comments


facchinm:
_Imported from https://github.com/arduino/Arduino/issues/6042 by @trivalik_ 

If we add a board and you want to add a comment after maximum_size entry an error occure.
Example for a line which will create an error:
`esplora.upload.maximum_size=130048 #131072`
The error is:
`strconv.ParseInt: parsing "130048 #131072": invalid syntax`

cmaglie:
I don't know, it may turns out that `#` is useful in some exotic command line syntax, I would like to not lose this possibility, so I'm closing this as wontfix.

PR #220, opened 2017-03-17
Allow inline comments in txt files

facchinm:
Fixes https://github.com/arduino/arduino-builder/issues/218


cmaglie:
The properties parser has now been moved to: https://github.com/arduino/go-properties-map

This PR should be moved there eventually.

PR #223, opened 2017-03-24
Add support for zipped sketch with embedded libraries

facchinm:


matthijskooijman:
Would this be a good moment to consider allowing a `libraries` subdirectory in all sketches? I think this PR adds support for this only for zipped sketches, which I would think is:
 1. Incosistent and leads to surprising behaviour
 2. A missed opportunity: Loading libraries from inside the sketch folder can greatly enhance the workflow of a complex sketch, for example when its dependency versions are kept using version control and git submodules.

lmihalkovic:
Cool ... i see you have played with the idea... the IDE side is pretty straightforward .. i started doing the building blocks again on the 1.9.0 a couple days ago.

lmihalkovic:
Using your suffix instead of mine (have not written support for the embedded libraries yet)

screen shot 2018-05-02 at 9 03 28 pm

screen shot 2018-05-02 at 9 02 57 pm

Issue #230, opened 2017-06-02
Include scanning cache broken for non-arduino boards


majenkotech:
First compilation works perfectly. Second compilation crashes arduino-builder:

```
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0x556bc0, 0xc420018130)
	/home/jenkins/go/src/runtime/panic.go:500 +0x1a1
arduino.cc/builder.findIncludesUntilDone(0xc420060400, 0xc4202f5d40, 0x556f40, 0xc4201f90e0, 0xc420381162, 0x18, 0x0, 0x0)
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/builder/container_find_includes.go:321 +0x138f
arduino.cc/builder.(*ContainerFindIncludes).Run(0x64d248, 0xc420060400, 0x41bc3c3c, 0x0)
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/builder/container_find_includes.go:149 +0x5e4
arduino.cc/builder.runCommands(0xc420060400, 0xc42019fad8, 0x22, 0x22, 0x1, 0x0, 0x0)
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/builder/builder.go:191 +0xcd
arduino.cc/builder.(*Builder).Run(0xc42019fd28, 0xc420060400, 0xc420045d88, 0x562240)
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/builder/builder.go:124 +0xb9c
arduino.cc/builder.RunBuilder(0xc420060400, 0xc420045d88, 0x64d248)
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/builder/builder.go:222 +0x35
main.main()
	/home/jenkins/workspace/arduino-builder-all-cross/src/arduino.cc/arduino-builder/main.go:338 +0x99c
/home/matt/Downloads/arduino-1.8.2/arduino-builder returned 2
```

Environment
-----------------

Core URL: https://raw.githubusercontent.com/chipKIT32/chipKIT-core/master/package_chipkit_index.json
Test board: chipKIT MAX32
Operating System: Linux (64 bit)
IDE version: 1.8.2

Steps to reproduce
-------------------------

1. Install the required boards package and select the right board
2. Install the required library (below)
3. Paste in the test sketch and compile (succeeds)
4. Compile again - fails.

Test Sketch
---------------

[...]

PR #236, opened 2017-06-16
Include cache fixes and refactorings

matthijskooijman:
This PR fixes a few corner cases in the include caching and refactors some code in the process. In particular, this fixes the issue reported at #230. The problem in that report consisted of three parts:

 - `arduino-builder` would fail to remove the `-MMD` option from commandline for the chipKIT32 boards, due to the way their `platform.txt` was structured. This caused the preprocessor command to fail, trying to write to `/dev/null.d`.
 - The preprocessor command would be run a second time to show an error, but with slightly different options, which prevented the error from occuring. This caused `arduino-builder` to ignore the failure and continue, masking the above problem.
 - Because the include cache would now be in a half-finished state, a small bug in the include caching was triggered that caused an out-of-bounds array read and a panic.

This PR fixes all three problems (in reverse order), with a fair amount of refactoring in between (some of which is strictly needed, some of which is "while I'm here").


matthijskooijman:
> Anyway, a lot of functions turned into using the global context; while the new signatures are slightly more readable, unit-testing could be much more convoluted but I don't have a strong opinion about it.

The policy I'm using here is that these helper functions can access things like ctx.DebugLevel and ctx.Verbose, which are essentially just global flags, but should still use explicit arguments for things like the list of files to compiler, or the buildProperties to use (even when these are also available through the context). In the long run, I think the latter should actually be removed from the context and be explicitely passed from function to function instead, to reduce the amount of implicit dataflow.

matthijskooijman:
@facchinm pointed out [some benchmarks](https://github.com/arduino/arduino-builder/pull/242#issuecomment-322320069) that show this PR significantly slows down compilation. This needs some investigation before merging this PR.

matthijskooijman:
I can't really reproduce the slow benchmarks of #242. I tried PR232, this PR, and this PR after rebasing, which all give similar times (there's a bit of variation, but I was using my computer while running these texts, so that probably explains the small differences).

```
                                                PR232    PR236    PR236 rebased
TestTryBuild001   ok    arduino.cc/builder/test 2.398s   2.103s   2.281s
TestTryBuild002   ok    arduino.cc/builder/test 2.081s   2.022s   1.977s
TestTryBuild003   ok    arduino.cc/builder/test 0.124s   0.133s   0.129s
TestTryBuild004   ok    arduino.cc/builder/test 1.997s   2.018s   1.982s
TestTryBuild005   ok    arduino.cc/builder/test 2.275s   2.162s   2.236s
TestTryBuild006   ok    arduino.cc/builder/test 2.105s   1.979s   2.094s
TestTryBuild007   ok    arduino.cc/builder/test 3.393s   3.375s   3.391s
TestTryBuild009   ok    arduino.cc/builder/test 2.039s   2.038s   2.030s
TestTryBuild010   ok    arduino.cc/builder/test 2.083s   2.038s   1.954s
TestTryBuild011   ok    arduino.cc/builder/test 2.008s   2.051s   2.043s
TestTryBuild012   ok    arduino.cc/builder/test 2.012s   2.043s   2.043s
TestTryBuild013   ok    arduino.cc/builder/test 2.066s   2.005s   2.077s
TestTryBuild014   ok    arduino.cc/builder/test 2.008s   2.006s   2.049s
TestTryBuild015   ok    arduino.cc/builder/test 2.351s   2.238s   2.287s
TestTryBuild017   ok    arduino.cc/builder/test 0.206s   0.320s   0.173s
TestTryBuild018   ok    arduino.cc/builder/test 2.242s   3.001s   2.258s
TestTryBuild019   ok    arduino.cc/builder/test 2.020s   2.626s   1.976s
TestTryBuild020   ok    arduino.cc/builder/test 0.221s   0.316s   0.221s
TestTryBuild021   ok    arduino.cc/builder/test 2.105s   2.370s   2.003s
TestTryBuild022   ok    arduino.cc/builder/test 9.257s   10.876s  9.241s
TestTryBuild023   ok    arduino.cc/builder/test 2.023s   2.429s   2.303s
TestTryBuild024   ok    arduino.cc/builder/test 1.999s   2.000s   2.627s
TestTryBuild026   ok    arduino.cc/builder/test 2.578s   2.655s   3.495s
TestTryBuild027   ok    arduino.cc/builder/test 2.307s   2.305s   2.600s
TestTryBuild028   ok    arduino.cc/builder/test 2.051s   2.160s   2.429s
TestTryBuild029   ok    arduino.cc/builder/test 2.069s   1.980s   2.721s
TestTryBuild030   ok    arduino.cc/builder/test 1.982s   2.191s   2.493s
TestTryBuild031   ok    arduino.cc/builder/test 3.461s   3.398s   4.421s
TestTryBuild032   ok    arduino.cc/builder/test 2.130s   2.211s   2.129s
TestTryBuild033   ok    arduino.cc/builder/test 1.960s   2.044s   1.999s
TestTryBuild034   ok    arduino.cc/builder/test 1.990s   2.039s   2.055s
TestTryBuild035   ok    arduino.cc/builder/test 1.973s   1.963s   2.007s
TestTryBuild036   ok    arduino.cc/builder/test 8.696s   8.847s   9.243s
TestTryBuild037   ok    arduino.cc/builder/test 2.003s   1.989s   2.010s
TestTryBuild038   ok    arduino.cc/builder/test 2.008s   1.978s   2.002s
TestTryBuild039   ok    arduino.cc/builder/test 5.239s   5.788s   5.368s
TestTryBuild040   ok    arduino.cc/builder/test 2.017s   2.588s   1.969s
TestTryBuild041   ok    arduino.cc/builder/test 1.990s   2.696s   1.977s
TestTryBuild042   ok    arduino.cc/builder/test 1.973s   2.478s   1.960s
```

For testing, I used ` for i in TestTryBuild{001..042}; do echo -ne "$i   "; go test arduino.cc/builder/test -run $i;done`

@facchinm, So it seems this PR can be merged? I'll ask in the other PR to see if those benchmarks can be reproduced, just in case, but I don't think this should block this PR any longer. I've also just rebased this PR on top of master, and re-ran the test suite, which still works.

matthijskooijman:
I also added three more cleanup commits from #217 (since that PR isn't quite ready, but those cleanup commits are fine to merge already).

matthijskooijman:
@facchinm, it indeed seems like the slowness of this PR was a one-off glitch, see https://github.com/arduino/arduino-builder/pull/242#issuecomment-349634584

matthijskooijman:
@facchinm @cmaglie, I just rebased this PR again, so it applies cleanly now. Could we perhaps merge this? Any issues were resolved, and since it touches quite some code, it is likely to run into new conflicts if we leave this unmerged.

cmaglie:
Everything has been manually merged.

matthijskooijman:
Thanks!

Issue #244, opened 2017-08-27
Can't install


webdev23:
Seriously whoever say that everyone know how to manage go-lang ... should go back to his java paradigm, definetly.

Please clarify your docs, mostly for the installation procedures, or better, make a one-click package installer.
There is hundreds of ways to make it simple, so why don't do it?

> 
>  go build arduino.cc/arduino-builder
> can't load package: package arduino.cc/arduino-builder: cannot find package "arduino.cc/arduino-builder" in any of:
> 	/snap/go/727/src/arduino.cc/arduino-builder (from $GOROOT)
> 	/home/zen/go/src/arduino.cc/arduino-builder (from $GOPATH)

matthijskooijman:
Not sure what the problem is exactly. If I follow the instructions (except that I do not run setup_go_vars entirely, just set GOPATH, we should probably improve that part), it builds correctly:

```
matthijs@grubby:/tmp$ git clone https://github.com/arduino/arduino-builder.git
Cloning into 'arduino-builder'...
remote: Counting objects: 4953, done.
remote: Total 4953 (delta 0), reused 0 (delta 0), pack-reused 4953
Receiving objects: 100% (4953/4953), 1.14 MiB | 977.00 KiB/s, done.
Resolving deltas: 100% (3131/3131), done.
matthijs@grubby:/tmp$ cd ar=
-bash: cd: ar=: No such file or directory
matthijs@grubby:/tmp$ cd arduino-builder/
matthijs@grubby:/tmp/arduino-builder$ export GOPATH=$(pwd)
matthijs@grubby:/tmp/arduino-builder$ go get github.com/go-errors/errors
matthijs@grubby:/tmp/arduino-builder$ go get github.com/stretchr/testify
matthijs@grubby:/tmp/arduino-builder$ go get github.com/jstemmer/go-junit-report
matthijs@grubby:/tmp/arduino-builder$ go build arduino.cc/arduino-builder
matthijs@grubby:/tmp/arduino-builder$ ls
arduino-builder  bin  CONTRIBUTING.md  fmt_fix_vet  LICENSE.txt  pkg  README.md  setup_go_env_vars  src
```

You'll have to provide more details about how you set things up. Also, I would suggest you reconsider your (even slightly) snarky tone, which seems inappropriate to me.

webdev23:
You are right. Please forgive. 
The procedure was failing without the  **export GOPATH=$(pwd)** line that you just shown.
[...]

PR #245, opened 2017-08-29
Improve build instructions in README

matthijskooijman:
This drops the `setup_go_env_vars` script, which is only confusing. It
contained three lines, of which only one (setting `GOPATH`) is always
required and the other two are only relevant when go is manually
installed. In the most common case (go installed through a package
manager), they will even break the build.

Instead of referring to an external script, just show how to set
`GOPATH` in the README itself.

This improvement was prompted by the user confusion found in #244.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,18 +57,22 @@ See [Doing continuous integration with arduino builder](https://github.com/ardui
 
 You need [Go 1.6.3](https://golang.org/dl/#go1.6.3).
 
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Since this is a fairly trivial change, that becomes useful as soon as it is merged (not only after a release), perhaps this can be merged already? @facchinm ?

facchinm:
Totally agree, merging right now

PR #264, opened 2018-02-19
Make `go build` work as intended

kolyshkin:
The aim of this commit is to make it so that

	go get github.com/arduino/arduino-builder

will build a binary named `arduino-builder` and put it into
`$GOPATH/bin` so it will be instantly available (provided
the user have `$GOPATH/bin` in their `$PATH`).

This is as easy as
1. move all go files to /builder subdir
2. move the main.go to a top dir
3. fix the path to builder in main.go and tests

[v2: fix tests]
[v3: fix README and travis.yml]


cmaglie:
You can already do the same with two `arduino-builder` in the path:

`go build github.com/arduino/arduino-builder/arduino-builder`

as explained in the readme: https://github.com/arduino/arduino-builder#building-from-source

kolyshkin:
> You can already do the same with two arduino-builder in the path:
>
> go build github.com/arduino/arduino-builder/arduino-builder
> 
> as explained in the readme: https://github.com/arduino/arduino-builder#building-from-source

I followed these instructions before attempting this PR. The problems are:
1. `arduino-builder` is not installed into `$GOPATH/bin` (as it's expected in such cases)
2. A few commands are needed to make it work, while after this PR it's just one (as you can see in the patch, check the first change to README.md).

matthijskooijman:
IIUC, the problem with the current package is that `main.go` lives in a subpackage (or however it's called in go), and not in the main `github.com/arduino/arduino-builder` package? If so, wouldn't it work to do `go get github.com/arduino/arduino-builder/arduino-builder`? Or doesn't that pull in the dependencies from the main repo or something?

In any case, the changes proposed in this PR seem good to me, I've always thought this "duplicate" `arduino-builder` name is a bit weird.

Another thought: Instead of moving all the .go files from the root into a `builder` directory (where 147f3d979b8da16b500a3fca49ce0f07476dd338 moved them out of recently), why not just only move `main.go` into the root? Wouldn't that also work?

kolyshkin:
>  wouldn't it work to do `go get github.com/arduino/arduino-builder/arduino-builder`?

In fact it works as intended! Not sure why I missed it when preparing this PR.

I'll close this one and prepare another -- for the sake of clarity.

> Another thought: Instead of moving all the .go files from the root into a builder directory (where 147f3d9 > moved them out of recently), why not just only move main.go into the root? Wouldn't that also work?

Unfortunately not -- go does not allow one to mix two packages in the same dir:

```
kir@kd:~/go/src/github.com/arduino/arduino-builder$ mv arduino-builder/main.go .
kir@kd:~/go/src/github.com/arduino/arduino-builder$ go build .
can't load package: package github.com/arduino/arduino-builder: found packages builder (add_additional_entries_to_context.go) and main (main.go) in /home/kir/go/src/github.com/arduino/arduino-builder
```

Yet another workaround would be to move the cli utility (i.e. main.go) to a separate package. Say, rename this repo from `arduino-builder` to just `builder` and then create `arduino-builder` repo, moving the `main.go` to it. This way it will be less renames, but one more repository.

PR #277, opened 2018-06-05
Remove composite paths support from foldersFlag

facchinm:
Fixes https://github.com/arduino/arduino-builder/issues/276
The only supported way to append multiple hardware/tools folders becomes specifying the same flag multiple times (which is the de-facto standard, since the Java IDE only uses this format)
@matthijskooijman

### DIFF
--- a/arduino-builder/main.go
+++ b/arduino-builder/main.go
@@ -96,19 +96,8 @@ func (h *foldersFlag) String() string {
 	return fmt.Sprint(*h)
 }
 
-func (h *foldersFlag) Set(csv string) error {
-	var values []string
-	if strings.Contains(csv, string(os.PathListSeparator)) {
-		values = strings.Split(csv, string(os.PathListSeparator))
-	} else {
-		values = strings.Split(csv, ",")
-	}
-
-	for _, value := range values {
-		value = strings.TrimSpace(value)
-		*h = append(*h, value)
-	}
-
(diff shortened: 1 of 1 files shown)


arduino-builder/main.go
```diff
@@ -97,17 +97,8 @@ func (h *foldersFlag) String() string {
 }
 
 func (h *foldersFlag) Set(csv string) error {
```
matthijskooijman: The `csv` variable should be renamed to `folder`.

arduino-builder/main.go
```diff
@@ -97,17 +97,8 @@ func (h *foldersFlag) String() string {
...
-	} else {
-		values = strings.Split(csv, ",")
-	}
-
-	for _, value := range values {
-		value = strings.TrimSpace(value)
-		*h = append(*h, value)
-	}
+	csv = strings.TrimSpace(csv)
+	*h = append(*h, csv)
```
matthijskooijman: Should we also remove the TrimSpace here? With it, you cannot have use directory name ending or starting in a space (which is unlikely to be useful, but why not just allow *any* filename?). A shell will trim spaces anyway, unless someone takes the effort to escape them (in which case we shouldn't trim them again)?

facchinm:
@matthijskooijman you are right about the rename, also the second one looks like an artificial limitation but we have to make sure that removing the trim brings no side effect

PR #288, opened 2018-09-04
Improve core archive info

facchinm:


phases/core_builder.go
```diff
@@ -124,10 +124,14 @@ func compileCore(ctx *types.Context, buildPath string, buildCachePath string, bu
 
 	// archive core.a
 	if targetArchivedCore != "" {
+		err := builder_utils.CopyFile(archiveFile, targetArchivedCore)
 		if ctx.Verbose {
-			logger.Println(constants.LOG_LEVEL_INFO, constants.MSG_ARCHIVING_CORE_CACHE, targetArchivedCore)
+			if err == nil {
+				logger.Println(constants.LOG_LEVEL_INFO, constants.MSG_ARCHIVING_CORE_CACHE, targetArchivedCore)
+			} else {
+				logger.Println(constants.LOG_LEVEL_INFO, constants.MSG_CORE_CACHE_UNAVAILABLE, ctx.ActualPlatform.PlatformId)
```
matthijskooijman: This blames *any* error while copying the file on the core maintainers. I suppose the intended case is where the core does not use the proper .a filename, so the copy cannot find it? If so, then perhaps a separate check should be made to see if the .a filename exists and give this error if it does not? If the file *does* exist, but the copy fails, a new error message should probably be shown?

builder_utils/utils.go
```diff
@@ -558,5 +558,10 @@ func GetCachedCoreArchiveFileName(fqbn, coreFolder string) string {
 		coreFolder = absCoreFolder
 	} // silently continue if absolute path can't be detected
 	hash := utils.MD5Sum([]byte(coreFolder))
-	return "core_" + fqbnToUnderscore + "_" + hash + ".a"
+	realName := "core_" + fqbnToUnderscore + "_" + hash + ".a"
+	if len(realName) > 100 {
+		// avoid really long names, simply hash the final part
+		realName = "core_" + utils.MD5Sum([]byte(fqbnToUnderscore+"_"+hash)) + ".a"
```
matthijskooijman: This feels a bit arbitrary, but it would probably work well in practice.

phases/core_builder.go
facchinm: I'd also add `arduino.ar` as a possible target but this clashes badly with specifications at http://goo.gl/QdCUjo .
I think that adding this check will make life terrible for maintainers of "incompatible" cores for a bit but will enormously speedup the adoption of proper naming (esp8266 just merged it in mainline https://github.com/esp8266/Arduino/commit/85e68093e94b4e5c5e711888b8b5facee1bd4f9f#diff-d638a47d1ca0e2a5585fb34c7d7fb0f9L95 )

builder_utils/utils.go
facchinm: Really arbitrary indeed; it still fails if someone has a TEMP directory with ~140 characters (`260 - 100 - len("arduino-core-cache-xxxxxx")`) but on Windows it would mean that your username is at least 100 keystrokes long (very unlikely)

phases/core_builder.go
matthijskooijman: Terrible in what way? Because users will start pestering them? Or because their incompatible cores will break or become slower (I can't see why or how)?

facchinm: The first one :slightly_smiling_face:

PR #295, opened 2018-09-20
Improve hex merger

### LINKED ISSUE #286: with_bootloader issues
The "with_bootloader" export which happens here...

https://github.com/arduino/arduino-builder/blob/29cf4c8b1473b4d0ba429c67ee497e2e061bd2fc/merge_sketch_with_bootloader.go

has some issues.

  1. There is no output or even logging that it happened, so we can't for example see which bootloader it picked easily, how much it threw away, and that sort of thing.

  2. The attached zip contains a compiled sketch .hex, the bootloader which it (in theory) selected, and the merged hex.  The merged hex however contains the sketch hex, and two lines only from the bootloader hex - you will see clearly it can't be right because the bootloader is 9.9k, and the "merged" file which should contain both the sketch and bootloader is only 640 bytes.

facchinm:
Fixes #286

WARNING: still doesn't work with Caterina based bootloaders (because of sections overlapping)


merge_sketch_with_bootloader.go
```diff
@@ -80,70 +83,85 @@ func (s *MergeSketchWithBootloader) Run(ctx *types.Context) error {
...
+		return err
 	}
+	defer mergeFile.Close()
+
+	mem_merge.DumpIntelHex(mergeFile, 16)
+
+	mergedSketchPathBin := strings.TrimSuffix(mergedSketchPath, ".hex") + ".bin"
 
-	return utils.WriteFile(mergedSketchPath, strings.Join(sketch, "\n"))
+	bytes := mem_merge.ToBinary(initial_address, last_address-initial_address, 0xFF)
```
facchinm: This must only be executed if address space is continuous (otherwise we can get a huge file as result)

facchinm:
Replaced by https://github.com/arduino/arduino-cli/pull/744

PR #299, opened 2018-10-05
Ensure that no spurious file is added to AdditionalFiles

facchinm:
If the build is executed "in place" (inside the sketch folder) intermediate cpp files were added to the AdditionaFiles list.
This was solved (badly) in https://github.com/arduino/arduino-builder/commit/849faa1f40336d51c9b16726213619e3d01cc19d#diff-b54df391d234f64238253f1d968e2e39R89 .

This patch fixes the issue by not including these files in the list in the first place, so we can avoid removing them later.

### DIFF
--- a/sketch_loader.go
+++ b/sketch_loader.go
@@ -76,7 +76,7 @@ func (s *SketchLoader) Run(ctx *types.Context) error {
 		return i18n.ErrorfWithLogger(logger, constants.MSG_CANT_FIND_SKETCH_IN_PATH, sketchLocation, filepath.Dir(sketchLocation))
 	}
 
-	sketch, err := makeSketch(sketchLocation, allSketchFilePaths, logger)
+	sketch, err := makeSketch(sketchLocation, allSketchFilePaths, ctx.BuildPath, logger)
 	if err != nil {
 		return i18n.WrapError(err)
 	}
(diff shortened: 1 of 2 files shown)


sketch_loader.go
```diff
@@ -123,7 +123,9 @@ func makeSketch(sketchLocation string, allSketchFilePaths []string, logger i18n.
 				otherSketchFiles = append(otherSketchFiles, sketchFile)
 			}
 		} else if ADDITIONAL_FILE_VALID_EXTENSIONS[ext] {
-			additionalFiles = append(additionalFiles, sketchFile)
+			if !strings.Contains(filepath.Dir(sketchFile.Name), buildLocation) {
+				additionalFiles = append(additionalFiles, sketchFile)
+			}
```
matthijskooijman: I believe this is not necessary, as recursive compilation is already limited to the `src/` subdirectory. Have you confirmed that this check ever returns false with a build dir inside the sketch dir?

facchinm: I checked and if I only remove the other line all temporary cpp files get added top the list (even if not in src)

facchinm: There's also a test that should not pass if files outside `src` were not added to the list (https://github.com/arduino/arduino-builder/blob/master/test/sketch_loader_test.go#L129) . Is it intended @cmaglie ?

cmaglie: If the test is there it probably means something. you should go back to the commit that added those lines and check why they have been added...

cmaglie: This is the commit that added the "extra" files to the test: 9d7d1d7cb95c12e73e9dd44df132a643089c3f40

facchinm:
Merged manually

PR #307, opened 2018-11-21
Handle reference files when expanding commandline

facchinm:
This commit allows to specify referenced file in your boards.txt with the notation $filename.
gcc accepts files containing arguments as @filename, but it normally applies only to defines or similar flags.
With the new notation, these files can contain include paths, which must be specified absolutely and change between every installation.
At runtime, the file content is being read and written ion a writable location with the actual resolved paths.

### DIFF
--- a/builder_utils/utils.go
+++ b/builder_utils/utils.go
@@ -31,9 +31,11 @@ package builder_utils
 
 import (
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I'm not sure I understand the point of this commit. Could you provide an example?

Where you write "board.properties", did you mean board.txt?

facchinm:
@matthijskooijman I fixed the typo, thanks.
An example could be an architecture with hundreds of include paths; instead than passing every path in the commandline, with this patch we can have
```
mycoolboard.compiler.cxxflags=@{build.variant.path}/cxxflags.txt
mycoolboard.compiler.includes=${build.variant.path}/includes.txt
```
```
recipe.cpp.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.c.flags} -DARDUINO={runtime.ide.version} -DARDUINO_{build.board} -DARDUINO_ARCH_{build.arch}  {build.extra_flags} {compiler.c.extra_flags} "-I{build.core.path}/api/deprecated" {includes} {compiler.cxxflags} {compiler.includes} -o "{object_file}" "{source_file}"
```

`cxxflags.txt` contains something like
```
-Wvla
-fno-rtti
-std=gnu++98
-Os
-Wall
-Wextra
-Wno-missing-field-initializers
-Wno-unused-parameter
-c
-fdata-sections
-ffunction-sections
-fmessage-length=0
-fno-builtin
-fno-delete-null-pointer-checks
-fno-exceptions
-fomit-frame-pointer
-funsigned-char
-g1
```
while `includes.txt` contains
```
-I{build.core.path}/superlongpath
-I{build.core.path}/anothersuperlongpath
-I{build.core.path}/and_so_on
...
```

During commandline creation, `$includes.txt` gets copied to build folder, resolved and used in the commandline as `@includes.txt`

matthijskooijman:
Ah, I see. So the point is to apply `{var}` substitution on an arbitrary file and pass that to the compiler.

I can see the usecase, but I'm not sure I like the implementation approach much. It seems very specific to the gcc @-syntax, while there might be other compilers or option that would benefit from the same. Also, it limits using the $ in recipes, requiring extra escaping and possibly increasing user surprise (as well as breaking compatibility with recipes that already use $ in an incompatible way).

Thinking about alternatives, could this long list of includes not be listed in platform.txt? Then you won't have the flags in the variant directory, but you could still have something in your platform.txt like:

    standard_includes=-I{build.core.path}/superlongpath
    standard_includes={standard_includes} -I{build.core.path}/anothersuperlongpath

(not sure if spreading a variable over multiple lines like this works, or is the best approach, though)

And in boards.txt:

    mycoolboard.compiler.includes={standard_includes}


That does mean you cannot use the `variant` dynamically and have to list the name of the includes variable for each board separately.

Alternatively, an "include another .txt file" could perhaps also solve this, since the variables in the .txt file are then again interpolated. This could be a lot more flexible. E.g. in boards.txt:

    mycoolboard.include_txt=${build.variant.path}/includes.txt

Then in that includes.txt:

    compiler.includes=-I{build.core.path}/superlongpath
    compiler.includes={compiler.includes} -I{build.core.path}/anothersuperlongpath

And the recipe can then just reference `compiler.includes` like in your example.

facchinm:
I agree with your approach for the most part, but beside paths being long in `includes.txt`, it also has ~250 lines :slightly_smiling_face: 
This strongly reduces the maintainability of `boards.txt`, beside surpassing by far the 30000 chars max commandline size.
`$stuff` is only expanded/mangled if `stuff` is a valid (and readable) file path, otherwise it's left unchanged. The behaviour is very gcc (and probably clang) specific, but doesn't have any side effect if unused.

facchinm:
Any alternative idea is well accepted :smile: 

The problem I'm trying to solve can only be tackled by NOT expanding all the stuff to the commandline (which would be the most sensible and generic approach) to keep its length under control. 
This limits a bit the possible approaches, apart from adding a "pre" rule like
```
compiler.includes=includes.txt
compiler.includes.path={build.variant.path}/{compiler.includes}
recipe.hooks.sketch.prebuild.1.pattern={runtime.tool.make_abs_file} -i {compiler.includes.path} -b {build.path}
```
which would create `{build.path}/preproc/{compiler.includes}` so it can be used in the compile pattern as  
```
recipe.cpp.o.pattern=.... @{build.path}/preproc/{compiler.includes} .... 
```

This means adding a specific tool to accomplish this, which I would prefer to avoid since I think the generic approach would improve some cores' maintainability :slightly_smiling_face:

matthijskooijman:
I wonder if we can somehow internalize this prebuild thing. Something like:
```
recipe.hooks.sketch.prebuild.1.type=interpolate
recipe.hooks.sketch.prebuild.1.input={compiler.includes.path}
recipe.hooks.sketch.prebuild.1.output={build.path}/preproc/{compiler.includes}
```

Issue #235, opened 2017-06-15
Setting build-path to nested subdirectory


GeorgeHahn:
Related to #4 - if the build dir is set to a folder within the working directory (eg `./build`), a recursive folder structure is created (see #4 for an example).

Issue #4 notes that an error message was added to guard against this behavior. This error message should be extended to cover any folder contained within the working directory.

matthijskooijman:
Hm, I thought we already fixed this and allowed a build directory within the sketch directory. If not, then it's certainly on my list of things that should be made possible :-)

luni64:
Any news on that? It still get an error when I define the build directory in the sketch directory
`Rel: can't make C:\Users\lutz\source\newProject(1)\bin\core\core.a relative to bin`
Any workaround?

facchinm:
@luni64 would you mind attaching the full compiler output? The problem looks a bit Windows specific

facchinm:
Thanks for the log.
I think there might be a workaround by using an absolute path for `-build-path=` (I would write `$PWD/bin` on Linux). The piece of code that kicks in is https://github.com/arduino/arduino-builder/blob/bbf705133803b0825818af1ed51cf694eef17fc7/phases/linker.go#L58-L61
For some reason `ctx.BuildPath` is still relative in that snippet; I'll take a look tomorrow to understand what is happening :slightly_smiling_face:

facchinm:
@luni64 I just pushed a test PR here https://github.com/arduino/arduino-builder/pull/308 , as soon as the binary is ready it would be cool if you could test it in your environment!

luni64:
Sure, I'll test it as soon as it is available

luni64:
That didn't work, refuses to compile at all now. Here the log:
```
> Executing task: & 'C:/Program Files/arduino-1.8.5/arduino-builder' -verbose=1 -logger=human '-hardware= C:\Program Files\arduino-1.8.5\hardware' -build-path=bin '-tools=C:\Program Files\arduino-1.8.5\tools-builder' '-tools=C:\Program Files\arduino-1.8.5\hardware\tools\avr' -fqbn=teensy:avr:teensy31:usb=serial,speed=96,opt=o2std,keys=en-us src\main.cpp <

FindFirstFile C:\Users\lutz\source\newProject(1)\ C:\Program Files\arduino-1.8.5\hardware: Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch.
[...]

PR #308, opened 2018-11-21
Make build path absolute in whole context

facchinm:
Test PR for https://github.com/arduino/arduino-builder/issues/235

### DIFF
--- a/arduino-builder/main.go
+++ b/arduino-builder/main.go
@@ -37,6 +37,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I had a look at this, I found that the core build path (and thus the core archive path) is made absolute here: https://github.com/arduino/arduino-builder/blob/master/add_additional_entries_to_context.go#L44

If the buildpath would be made absolute earlier, than all these filepath.Abs() could be removed, I guess.

If the path is made absolute, this is probably the best place: https://github.com/arduino/arduino-builder/blob/master/arduino-builder/main.go#L314

If no explicit build path is generated, an automatic one is generated which is probably already absolute: https://github.com/arduino/arduino-builder/blob/master/generate_buildpath_if_missing.go
https://github.com/arduino/arduino-builder/blob/master/generate_buildpath_if_missing.go

facchinm:
Agree with going back to the source https://github.com/arduino/arduino-builder/blob/bbf705133803b0825818af1ed51cf694eef17fc7/arduino-builder/main.go#L314 :slightly_smiling_face: Will modify the PR in a bit-

Issue #171, opened 2016-07-26
WARNING: Spurious .vscode folder in library


steronydh:
MS Visual Studio Code adds a .vscode folder when used as an external editor, which causes Arduino 1.6.9 to show the above warning.

Please can .vscode be added to the list of folders to ignore (.git etc. are already ignored).

PR #315, opened 2019-03-11
added .pioenvs and .piolibdeps to ignore list

aster94:
Done to avoid the error:
```
WARNING: Spurious *** folder in '***' library
```
the two folders cames from Platformio

someone could point out that it would be meaningless to add them since people who have this warning are using platformio and not the Arduino IDE, but really, IMHO, many of them to write fast and simple skecth will still use the Arduino IDE

similar to this PR:
https://github.com/arduino/arduino-builder/issues/70
https://github.com/arduino/arduino-builder/issues/171 fixed with https://github.com/arduino/arduino-builder/commit/d71753a9299f6f97cbef805b672d2ead83ee7bbe

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -152,7 +152,7 @@ func FilterFiles() filterFiles {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


aster94:
Thanks @facchinm

PR #331, opened 2019-08-26
[DO NOT MERGE] Fix parallel compilation disregarding -jobs flag

facchinm:
The issue was due to the peculiar way concurrency and parallelism are handled in go.
We used to set GOMAXPROC to 1 to avoid parallelizing the WaitGroup execution.
This would work, in theory, unless the goroutines sleep.
In that case, another goroutine is allowed to start concurrently (only 1 goroutine running in parallel, so GOMAXPROC is happy).
Since our goroutines sleep (wait) after calling gcc, another task is started, without any hard limit, till the WaitGroup is completely spawned.
On systems with limited resources (as RaspberryPi) and cores with lots of files this can trigger an out of memory condition.

### DIFF
--- a/arduino-builder/main.go
+++ b/arduino-builder/main.go
@@ -207,13 +207,14 @@ func main() {
 		return
 	}
 
-	if *jobsFlag > 0 {
-		runtime.GOMAXPROCS(*jobsFlag)
-	} else {
-		runtime.GOMAXPROCS(runtime.NumCPU())
-	}
-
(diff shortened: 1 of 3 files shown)


facchinm:
@ArduinoBot build this please

facchinm:
@ArduinoBot build this please

cmaglie:
merged and fixed in https://github.com/arduino/arduino-cli/pull/356

